PR #422, opened 2024-02-14
Use ruff

### LINKED ISSUE #426: [Feature]: Use ruff for linting
### Describe the feature that you would like.*

Replace the existing linting and formatting tools, where possible, with their equivalents in ruff, and replace setuptools with hatchling.

### Why is this feature valuable to you?*

Ruff provides extremely fast linting, is actively supported, and easily configurable. In addition, it provides parity with some now deprecated linting tools such as [pydocstyle](https://github.com/pycqa/pydocstyle). Ruff also provides black-like formatting support (see here for [known deviations from black](https://docs.astral.sh/ruff/formatter/black/)), removing the need for an additional dependency.

The full list of benefits for moving the build system to [hatchling are listed here](https://hatch.pypa.io/latest/why/#build-backend).

### Describe alternatives you've considered

At present ruff does not currently provide support for all the [linting rules provided by pylint](https://docs.astral.sh/ruff/rules/#pylint-pl), all though most are available in "preview" mode. Once out of preview, it may be possible to remove the `.pylintrc` and rely solely on ruff for all linting and formatting needs.

### Additional context

_No response_

### Code of Conduct*

- [X] I agree to follow this project's Code of Conduct

tc85324:
### PR Type
Closes #426 
- Build related changes
- CI related changes

### Description
- Simplify and speed-up the pre-commit with [ruff](https://docs.astral.sh/ruff/).

Ideally .pylintrc could be replaced in the future once full support is provided in ruff?

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -103,7 +103,7 @@ ignore = [
     "E114",   # indentation-with-invalid-multiple-comment
     "E117",   # over-indented
     "W191",   # tab-indentation
-    # Incompatible with other pydocstyle rules.
+    # Incompatible with other pydocstyle rules
     "D203",   # one-blank-line-before-class
     "D212",   # multi-line-summary-first-line
     # Opinionated ignores
--- a/tests/unit/test_kernel.py
+++ b/tests/unit/test_kernel.py
@@ -34,7 +34,7 @@ import coreax.kernel
 
 class TestKernelABC(unittest.TestCase):
     """
-    Tests related to the Kernel abstract base class in `kernel.py`.
+    Tests related to the Kernel abstract base class in ``kernel.py``.
     """
 
     def test_invalid_init_inputs_valid(self) -> None:
@@ -77,7 +77,7 @@ class TestKernelABC(unittest.TestCase):
 
 class TestSquaredExponentialKernel(unittest.TestCase):
     """
-    Tests related to the SquaredExponentialKernel defined in `kernel.py`.
+    Tests related to the SquaredExponentialKernel defined in ``kernel.py``.
     """
 
     def test_squared_exponential_kernel_init(self) -> None:
@@ -697,7 +697,7 @@ class TestSquaredExponentialKernel(unittest.TestCase):
 
 class TestLaplacianKernel(unittest.TestCase):
     """
-    Tests related to the LaplacianKernel defined in `kernel.py`.
+    Tests related to the LaplacianKernel defined in ``kernel.py``.
     """
 
     def test_laplacian_kernel_init(self) -> None:
@@ -1157,7 +1157,7 @@ class TestLaplacianKernel(unittest.TestCase):
 
 class TestPCIMQKernel(unittest.TestCase):
     """
-    Tests related to the PCIMQKernel defined in `kernel.py`.
+    Tests related to the PCIMQKernel defined in ``kernel.py``.
     """
 
     def test_pcimq_kernel_init(self) -> None:
@@ -1443,7 +1443,7 @@ class TestPCIMQKernel(unittest.TestCase):
 
 class TestSteinKernel(unittest.TestCase):
     """
-    Tests related to the SteinKernel defined in `kernel.py`.
+    Tests related to the SteinKernel defined in ``kernel.py``.
     """
 
     def test_stein_kernel_computation(self) -> None:
--- a/coreax/metrics.py
+++ b/coreax/metrics.py
@@ -36,8 +36,6 @@ import coreax.util
 import coreax.validation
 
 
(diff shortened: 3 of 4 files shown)


coreax/score_matching.py
```diff
@@ -143,7 +142,7 @@ def _tree_flatten(self):
 
         return children, aux_data
 
-    def __init__(
+    def __init__(  # noqa: PLR0913, PLR0917
```
gchqdev227, line 145: I thought we had updated the max arguments in the pylint config?

examples/david_map_reduce_weighted.py
```diff
@@ -107,8 +107,9 @@ def main(
     image_data = cv2.cvtColor(original_data, cv2.COLOR_BGR2GRAY)
 
     print(f"Image dimensions: {image_data.shape}")
-    pre_coreset_data = np.column_stack(np.nonzero(image_data < 255))
-    pixel_values = image_data[image_data < 255]
+    max_8bit = 255
```
gchqdev227, line 110: This should be a module-level constant please!

pyproject.toml
```diff
@@ -75,16 +73,52 @@ Repository = "https://github.com/gchq/coreax"
...
+    "D",   # pydocstyle
+    "E",   # pycodestyle [error]
+    "F",   # pyflakes
+    "I",   # isort
+    "N",   # pep8-naming
+    "PL",  # pylint
+    "W"    # pycodestyle [warning]
+]
+ignore = [
+    "D203",
```
gchqdev227, line 99: I know we can look them up, but can you hint towards them in comments please?

tests/performance/test_kernel.py
```diff
@@ -47,7 +47,7 @@ def setUp(self) -> None:
 
 class TestSquaredExponentialKernel(TestKernel):
     """
-    Tests related to the SquaredExponentialKernel defined in kernel.py
+    Tests related to the SquaredExponentialKernel defined in `kernel.py`.
```
gchqdev227, line 50: These don't get rendered so it doesn't *really* matter, but for consistency please could they be in double back ticks?

tests/unit/test_kernel.py
```diff
@@ -951,7 +951,7 @@ def test_scaled_laplacian_kernel_gradients_wrt_x(self) -> None:
         .. math:
 
             - \text{output_scale}\operatorname{sgn}{(x - y)}{2length\_scale^{2}}e^{-\frac{\lVert x - y\rVert_1}{2 \text{length_scale}^2}}
-        """
+        """  # noqa: E501
```
gchqdev227, line 954: Line too long? Is there a way we can rewrite the LaTeX so that isn't a problem?

tests/unit/test_metrics.py
```diff
@@ -674,7 +674,7 @@ def test_mmd_weight_block_equals_mmd(self) -> None:
         r"""
         Test mmd_weight_block equals mmd when weights are uniform: :math:`w = 1/n`,
         :math:`w_y = 1/m`.
-        """
+        """  # noqa: D205
```
gchqdev227, line 677: Can we add in the blank line instead?

gchqdev227 (changes requested):
A few minor changes/discussion points, but otherwise great stuff!

coreax/score_matching.py
tc85324: We have, but I thought this specific case was on the side of unreasonable to be the limit for the whole project. I.E we have set a project wide value of 10, this case is 13/14.

tests/unit/test_kernel.py
tc85324: Yes. In the CONTRIBUTING.md we ask "Prioritise overfull lines for mathematical expressions over artificially splitting them into multiple equations in both comments and docstrings." so I think this should be left as is?

pyproject.toml
tc85324: Do you want hints in the places where noqa is used as well?

coreax/score_matching.py
tc85324: Oh I see what you mean. The values in the `.pylintrc` diverge from those I've set for ruff in `pyproject.toml`. I'll just set them to match and remove the noqa

tc85324:
The newest push includes changes to pylintrc (to align with the values chosen in pyproject.toml for ruff). In addition, it removes many of the subsequently redundant pylint enable/disable comments. Finally, fast_fail has been enabled for local use, and handling to disable fast_fail on the CI has been introduced.

tests/unit/test_kernel.py
gchqdev227: Fair point!

gchqdev227 (approved):
Looks good!

PR #449, opened 2024-02-19
[feat]: rename weights optimisers

### LINKED ISSUE #448: [Feature]: Give`coreax.metric.MMD` and `coreax.weights.MMD` unique names
### Describe the feature that you would like.*

There exists both `coreax.metric.MMD` and `coreax.weights.MMD`, but it is not clear to a newcomer that the following are not importing the same class:
```python
from coreax import MMD # coreax.metric.MMD
from coreax.weights import MMD # coreax.weights MMD
```

In the examples this requires:
```python
from coreax import MMD
from coreax.weights import MMD as MMDWeightsOptimiser
```

If the `coreax.weights.MMD` is renamed to `coreax.weights.MMDWeightsOptimiser` then any ambiguity is removed.

### Why is this feature valuable to you?*

Results in an easier to understand codebase.

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

### Code of Conduct*

- [X] I agree to follow this project's Code of Conduct

tc85324:
### PR Type
Closes #448 

- Any uses of `coreax.weights.MMD` will need to be replaced with `coreax.weights.MMDWeightsOptimiser`.
- Any uses of `coreax.weights.SBQ` will ned to be replaced with `coreax.weights.SBQWeightsOptimiser`.

### DIFF
--- a/coreax/weights.py
+++ b/coreax/weights.py
@@ -82,7 +82,7 @@ class WeightsOptimiser(ABC):
         return self.solve(x, y)
 
 
-class SBQ(WeightsOptimiser):
+class SBQWeightsOptimiser(WeightsOptimiser):
     """
     Define the Sequential Bayesian Quadrature (SBQ) optimiser class.
 
@@ -126,7 +126,7 @@ class SBQ(WeightsOptimiser):
         return jnp.linalg.solve(kernel_mm, kernel_nm)
 
 
-class MMD(WeightsOptimiser):
+class MMDWeightsOptimiser(WeightsOptimiser):
     r"""
     Define the MMD weights optimiser class.
 
--- a/coreax/weights.py
+++ b/coreax/weights.py
@@ -37,6 +37,7 @@ from abc import ABC, abstractmethod
 import jax.numpy as jnp
 from jax import Array
 from jax.typing import ArrayLike
+from typing_extensions import deprecated
 
 import coreax.kernel
 import coreax.util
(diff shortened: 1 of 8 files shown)


gchqdev227 (changes requested):
Overall this is a good change, but in order to fulfil our OSS obligations I would request that you add aliases for the `MMD` and `SBQ` classes. My suggestion would be to do something like:

```python
import warnings

class MMD(MMDWeightsOptimiser):

    def __init__(self, *args, **kwargs):
        warnings.warn("This class is now deprecated in favour of 'MMDWeightsOptimiser', and will be removed in version 0.3", category=DeprecationWarning)
        super().__init__(*args, **kwargs)
```
At some point we may wish to then to explicitly undocument these aliases, but for now it's not the end of the world if they make it in.

Also, please add an entry in the changlog!

tc85324:
Added changelog and deprecation warnings as requested.

coreax/weights.py
```diff
@@ -182,3 +183,19 @@ def solve(self, x: ArrayLike, y: ArrayLike, epsilon: float = 1e-10) -> Array:
         sol = coreax.util.solve_qp(kernel_mm, kernel_nm)
 
         return sol
+
+
+@deprecated("Renamed to SBQWeightsOptimiser; will be removed in version 0.3.0")
```
gchqdev227, line 188: Neat decorator use!!

gchqdev227 (approved):
All looks good, thanks for sorting that!

PR #450, opened 2024-02-19
Feature/improved docs

gchqdev227:
### PR Type

- Build related changes

### Description
Updated the documentation to be easier to navigate, and to fix build errors.  There are a few other changes I plan to make in this PR that have yet to be pushed, but since it's a large change I think it makes sense to start reviewing the first commit.

Summary of changes (in no particular order):

* The `automodule` calls have been split among multiple `.rst` files, and organised into more of a tree structure to ensure they display in a TOC on the left-hand side.
* Auto-conversion of markdown to rst has been replaced (it was causing too many headaches with headings and conversion). The examples have been typed up in rst and moved from the README (we agreed it was a more appropriate place). There is still some duplication across the docs and the README, but we can figure out where we would prefer everything to go over time.
* Example code blocks have been extracted to a `documentation/source/snippets` folder so that they can be subject to style checks.
* The documentation now contains the logo in the corner, and a favicon!
* The right-side TOC has been simplified to remove the full qualname of the methods, as we have unusually long method names that were hard to read.
* A tidy of `conf.py` to remove unused code. Also a refactor of functions that are still in use.
* Additions of footer icons, as well as a link at the top to view/edit the source code.
* Added custom bibtex style so that current bib keys can be used.
* Edited docstrings throughout code to fix incorrect internal links.
* Added `sphinx_toolbox.wikipedia` extension to allow simpler links to pages on Wikipedia.

### How Has This Been Tested?

The docs now build without warnings, meaning that we can eventually switch on the `fail_on_warnings` config option in `.readthedocs.yaml`.

No, (not to the code at least).

### Screenshots

n/a


README.md
```diff
@@ -1,5 +1,5 @@
-<div align="center">
-<img alt="Coreax logo" src="https://raw.githubusercontent.com/gchq/coreax/main/documentation/assets/Logo.svg">
+<div style="align: center;">
+    <img alt="Coreax logo" src="https://raw.githubusercontent.com/gchq/coreax/main/documentation/assets/Logo.svg">
 </div>
```
tc85324, line 3: While the "align" tag is deprecated, it is the only way that I'm aware of, to perform alignment in GitHub readmes. This is the result of sanitization that occurs behind the scenes. In the gaffer [README](https://github.com/gchq/Gaffer/blob/7f8832f1ed2ddd43286434fc19689d0654c9bb91/README.md?plain=1#L1) the align tag is used similarly to here. 

Alternatively, we could just make the logo smaller and have it left aligned?

coreax/coresubset.py
```diff
@@ -274,7 +273,8 @@ def _greedy_body(
             iteration ``i``.
         :param x: :math:`n \times d` data matrix
         :param kernel_vectorised: Vectorised kernel computation function. This should be
-            the :meth:`compute` method of a :class:`~coreax.kernel.Kernel` object,
+            the :meth:`~coreax.kernel.Kernel.compute` method of a
+            :class:`~coreax.kernel.Kernel` object,
```
tc85324, line 277: I thought the linter enforced no punctuation at the end of parameter descriptions, but maybe not. In any case, to be consistent, the trailing comma probably wants removing here.

coreax/reduction.py
```diff
@@ -97,6 +100,9 @@ def __init__(
             refine_method, "refine_method", (coreax.refine.Refine, type(None))
         )
         self.refine_method = refine_method
+        """
+        Refine method
+        """
```
tc85324, line 105: Are these required by the automatic documentation generator? Could we instead document the attributes in the class docstring?

tc85324 (commented):
Looks good to me. My comments are mostly just nitpicks.

.pylintrc
```diff
@@ -424,6 +424,7 @@ confidence=HIGH,
 disable=too-many-lines,
         too-many-arguments,
         too-many-instance-attributes,
+        import-error,
```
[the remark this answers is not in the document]
gchqdev227: My thinking was that this only gets raised when a file tries to import something that isn't installed, and if this is because we're actually missing a requirement then the test will catch them and fail anyway.  I could imagine a scenario in which a fresh repo is cloned and not all of the dev requirements are installed, and pylint would then complain!

README.md
gchqdev227: Ah right, fair enough! I will revert this back to what we had before as we know that it works!

coreax/kernel.py
```diff
@@ -1035,7 +1043,7 @@ def compute_elementwise(
 # Define the pytree node for the added class to ensure methods with JIT decorators
 # are able to run. This tuple must be updated when a new class object is defined.
 kernel_classes = (SquaredExponentialKernel, PCIMQKernel, SteinKernel, LaplacianKernel)
-for current_class in kernel_classes:
+for _current_class in kernel_classes:
     tree_util.register_pytree_node(
-        current_class, current_class.tree_flatten, current_class.tree_unflatten
+        _current_class, _current_class.tree_flatten, _current_class.tree_unflatten
     )
```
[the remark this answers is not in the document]
gchqdev227: I would prefer to do that in a separate PR as it feels like it will require careful re-testing.

coreax/reduction.py
gchqdev227: Some of these docstrings already existed, so I tried to use what we had already. Agree that putting it in the docstring is probably preferable to having random multiline strings under each assignment in `__init__`, but I added a placeholder one for each that is referenced from elsewhere in the code in order to silence warnings. It's worth noting that most of them aren't sufficiently descriptive and it might be worth either fleshing them out in a future ticket (which could also reformat them), or removing them (and their corresponding references) entirely.

tc85324 (approved):

coreax/reduction.py
tc85324: Ok, lets leave them for a future ticket then.

PR #499, opened 2024-03-05
Feature/input validation metrics 456

pc532627:
### PR Type

- Tests

### Description

Update input validation in metrics.py

### How Has This Been Tested?

New unit tests written and all pass on local machine.

N/A

### DIFF
--- a/tests/unit/test_metrics.py
+++ b/tests/unit/test_metrics.py
@@ -29,6 +29,8 @@ import coreax.kernel
 import coreax.metrics
 import coreax.util
 
+# pylint: disable=too-many-public-methods
+
 
 class TestMetrics(unittest.TestCase):
     r"""
@@ -119,7 +121,7 @@ class TestMMD(unittest.TestCase):
 
     def test_mmd_ones(self):
         r"""
-        Test mmd function with a small example dataset of ones and zeros.
+        Test MMD computation with a small example dataset of ones and zeros.
 
         For the dataset of 4 points in 2 dimensions, :math:`x`, and another dataset
         :math:`y`, given by:
@@ -176,7 +178,7 @@ class TestMMD(unittest.TestCase):
 
     def test_mmd_ints(self):
         r"""
-        Test MMD function with a small example dataset of integers.
+        Test MMD computation with a small example dataset of integers.
 
         For the dataset of 3 points in 2 dimensions, :math:`x`, and second dataset
         :math:`y`, given by:
@@ -235,7 +237,7 @@ class TestMMD(unittest.TestCase):
 
     def test_mmd_rand(self):
         r"""
-        Test that MMD computed from randomly generated test data agrees with mmd().
+        Test MMD computed from randomly generated test data agrees with method result.
         """
         # Define a kernel object
         length_scale = 1.0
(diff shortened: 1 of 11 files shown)


tests/unit/test_metrics.py
```diff
@@ -433,9 +436,84 @@ def test_sum_pairwise_distances_large_block_size(self) -> None:
...
+        # Define a kernel object and set pairwise computations to be the square distance
+        kernel = MagicMock()
+        kernel.compute = coreax.util.squared_distance_pairwise
+
+        # Define a metric object
+        metric = coreax.metrics.MMD(kernel=kernel)
+
+        # Compute sum of pairwise distances with a negative block_size - this should be
+        # capped at 0, and then the full text of the inbuilt error raised from the range
+        # function should provide all information needed for users to identify the issue
```
rg936672, line 481: Given that we already devote a line to capping block size at >=0, are we sure we don't want to also raise a warning/error if we do get a negative block size? Same comment on the other `_negative_block_size` methods

tests/unit/test_metrics.py
```diff
@@ -433,9 +436,84 @@ def test_sum_pairwise_distances_large_block_size(self) -> None:
...
+        """
+        Test sum_pairwise_distances when a negative block size is given.
+        """
+        # Setup some data
+        x = jnp.array([[0, 0], [1, 1], [2, 2]])
+        y = jnp.array([[0, 0], [1, 1]])
+
+        # Define a kernel object and set pairwise computations to be the square distance
+        kernel = MagicMock()
+        kernel.compute = coreax.util.squared_distance_pairwise
```
rg936672, line 474: We don't need to set `kernel.compute` if we're expecting to get an exception before it actually gets called

tests/unit/test_metrics.py
```diff
@@ -433,9 +436,84 @@ def test_sum_pairwise_distances_large_block_size(self) -> None:
...
         self.assertAlmostEqual(output, expected_output, places=5)
 
-    def test_mmd_block_ints(self) -> None:
+    def test_sum_pairwise_distances_zero_block_size(self):
+        """
+        Test sum_pairwise_distances when a zero block size is given.
+        """
+        # Setup some data
+        x = jnp.array([[0, 0], [1, 1], [2, 2]])
+        y = jnp.array([[0, 0], [1, 1]])
```
rg936672, line 445: Can't we just use the existing `self.x` and `self.y`? (Same comment on the rest of the added tests)

tests/unit/test_metrics.py
rg936672: (Same comment on the rest of the added tests)

tests/unit/test_metrics.py
pc532627: I've updated the code so that we have more clear error messages that flag this - not exactly a separate check for negative, but now not a cap at 0, but an error raised with clearer text

tests/unit/test_metrics.py
pc532627: Updated

tests/unit/test_metrics.py
```diff
@@ -433,9 +436,84 @@ def test_sum_pairwise_distances_large_block_size(self) -> None:
         # Check output matches expected
         self.assertAlmostEqual(output, expected_output, places=5)
 
-    def test_mmd_block_ints(self) -> None:
+    def test_sum_pairwise_distances_zero_block_size(self):
```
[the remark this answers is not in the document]
pc532627: I propose we leave this until a uniform style agreement has been reached - are you ok with that?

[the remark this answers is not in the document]
rg936672: Sounds sensible, yes

rg936672 (approved):

PR #503, opened 2024-03-06
Feature/input validation utils 461

pc532627:
### PR Type

### Description

Input validation in util.py refactored.

### How Has This Been Tested?

New tests added to test_utils.py, all pass on local machine.

N/A

### DIFF
--- a/coreax/util.py
+++ b/coreax/util.py
@@ -33,8 +33,6 @@ from jax import Array, jit, vmap
 from jax.typing import ArrayLike
 from jaxopt import OSQP
 
-import coreax.validation
-
 #: Kernel evaluation function.
 KernelComputeType = Callable[[ArrayLike, ArrayLike], Array]
 
@@ -70,23 +68,8 @@ def apply_negative_precision_threshold(
     :param precision_threshold: Positive threshold we compare against for precision
     :return: ``x``, rounded to 0.0 if it is between ``-precision_threshold`` and 0.0
     """
-    # Validate inputs
-    x = coreax.validation.cast_as_type(x=x, object_name="x", type_caster=float)
-    precision_threshold = coreax.validation.cast_as_type(
-        x=precision_threshold, object_name="precision_threshold", type_caster=float
-    )
-    coreax.validation.validate_in_range(
-        x=precision_threshold,
-        object_name="precision_threshold",
-        strict_inequalities=False,
-        lower_bound=0,
-    )
-
     if precision_threshold < 0.0:
-        raise ValueError(
-            f"precision_threshold must be positive; value {precision_threshold} given."
-        )
-
+        raise ValueError("precision_threshold must be positive.")
     if -precision_threshold < x < 0.0:
         return 0.0
 
@@ -103,9 +86,8 @@ def squared_distance(x: ArrayLike, y: ArrayLike) -> Array:
     :return: Dot product of ``x - y`` and ``x - y``, the square distance between ``x``
         and ``y``
     """
-    # Validate inputs
-    x = coreax.validation.cast_as_type(x=x, object_name="x", type_caster=jnp.atleast_1d)
-    y = coreax.validation.cast_as_type(x=y, object_name="y", type_caster=jnp.atleast_1d)
+    x = jnp.atleast_1d(x)
+    y = jnp.atleast_1d(y)
     return jnp.dot(x - y, x - y)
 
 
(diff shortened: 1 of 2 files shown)


coreax/util.py
```diff
@@ -186,20 +165,21 @@ def solve_qp(kernel_mm: ArrayLike, kernel_matrix_row_sum_mean: ArrayLike) -> Arr
...
-        expected_type=Array,
-    )
-
     # Setup optimisation problem - all variable names are consistent with the OSQP
-    # terminology. Begin with the objective parameters
-    q_array = jnp.array(kernel_mm)
-    c = -jnp.array(kernel_matrix_row_sum_mean)
+    # terminology. Begin with the objective parameters.
+    try:
+        q_array = jnp.array(kernel_mm)
```
tc85324, line 171: The intent is clearer if `jnp.asarray` is used instead.

coreax/util.py
```diff
@@ -186,20 +165,21 @@ def solve_qp(kernel_mm: ArrayLike, kernel_matrix_row_sum_mean: ArrayLike) -> Arr
...
+    except TypeError as exception:
+        raise TypeError(
+            "kernel_mm must be able to be converted to a JAX array"
+        ) from exception
+
+    try:
+        c = -jnp.array(kernel_matrix_row_sum_mean)
```
tc85324, line 178: As above, the intent is clearer if `jnp.asarray` is used instead.

tests/unit/test_util.py
```diff
@@ -140,7 +140,47 @@ def test_apply_negative_precision_threshold_valid_positive_input(self) -> None:
...
+
+        The output of solve_qp is indirectly tested when testing the various weight
+        optimisers that are used in this codebase. This test just ensures sensible
+        behaviour occurs when unexpected inputs are passed to the function.
+        """
+        # Attempt to solve a QP with an input that cannot be converted to a JAX array -
+        # this should error as no sensible result can be found in such a case.
+        with self.assertRaises(TypeError) as error_raised:
+            coreax.util.solve_qp(
+                kernel_mm="invalid_kernel_mm",
```
tc85324, line 155: Without the try excepts that have been added in the `coreax.util.solve_qp`, what happens in this case? Does solve_qp throw its own error?

coreax/util.py
```diff
@@ -70,23 +68,8 @@ def apply_negative_precision_threshold(
...
-        strict_inequalities=False,
-        lower_bound=0,
-    )
-
     if precision_threshold < 0.0:
-        raise ValueError(
-            f"precision_threshold must be positive; value {precision_threshold} given."
-        )
-
+        raise ValueError("precision_threshold must be positive.")
```
tc85324, line 72: To be precise, the check requires the `precision_threshold` to be non-negative (i.e. any positive value inclusive of zero), not positive. Depending on the intent, either the check or the error message should be changed.

tc85324 (changes requested):
My requested changes are very minor. I have one question regarding the unmittigated handling of invalid values but otherwise is good.

coreax/util.py
pc532627: Agreed and updated

coreax/util.py
pc532627: Agreed and updated

tests/unit/test_util.py
pc532627: The code ultimately returns an error of:

`TypeError: Value 'invalid_kernel_mm' with dtype <U17 is not a valid JAX array type. Only arrays of numeric types are supported by JAX.`

Which is reasonably helpful - the minor modification I made with the try/except logic was just to avoid the enormous error message that proceeds this - and instead highlight only the real issue present

coreax/util.py
pc532627: Thanks for catching this - I've updated the error text, the underlying code logic was the correct part

coreax/util.py
```diff
@@ -70,23 +68,8 @@ def apply_negative_precision_threshold(
...
-        strict_inequalities=False,
-        lower_bound=0,
-    )
-
     if precision_threshold < 0.0:
-        raise ValueError(
-            f"precision_threshold must be positive; value {precision_threshold} given."
-        )
-
+        raise ValueError("precision_threshold must not be negative.")
```
tc85324, line 72: Is there any reason not to use the f-string like error message from before?

coreax/util.py
```diff
@@ -70,23 +68,8 @@ def apply_negative_precision_threshold(
...
     if -precision_threshold < x < 0.0:
```
tc85324, line 73: The type hints don't align with the feasible logic here. For example:
```python
import jax.numpy as jnp
import coreax.util

x = jnp.array([-0.1, 1, 2])  # This is an ArrayLike
precision_threshold = 0.2

x_rounded = coreax.util.apply_negative_precision_threshold(x, precision_threshold)
# ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
The output type hint is also inconsistent. Ignoring the conditional statement, `x` will be returned identically to how it was passed, so the output is an `ArrayLike` (which is not necessarily a float).

tc85324 (changes requested):

coreax/util.py
pc532627: Thanks for catching this - the `ArrayLike` annotation on `x` was wrong - when the function was written, the docstring also states x as a scalar, I've updated the type annotation.

coreax/util.py
pc532627: Generally, avoiding printing user-inputs has been the approach taken

tc85324 (approved):

PR #507, opened 2024-03-12
fix: avoid merge conflicts if branch already exists

### LINKED ISSUE #505: Pre-commit autoupdate merge conflict
### What's the problem?

When the pre-commit autoupdate workflow runs, if .pre-commit-config.yaml differs between main and an existing chore/pre-commit-autoupdate branch, merge conflicts may occur and be committed unresolved. The result is an invalid .pre-commit-config.yaml.

1. Adjust a version to be very outdated in .pre-commit-config.yaml and commit to chore/pre-commit-autoupdate.
2. Wait for the weekly pre-commit autoupdate workflow to run.
3. See a merge conflict in .pre-commit-config.yaml like:

```
<<<<<<< chore/pre-commit-autoupdate
    rev: v8.4.0
=======
    rev: v8.6.0
>>>>>>> local
```

### Python version

3.9, 3.10, 3.11

### Package version

0.1.0

### Operating system

GitHub

### Other packages

_No response_

### Relevant log output

_No response_

tp832944:
closes #505

### PR Type

- Bugfix
- CI related changes

### Description

Copy the version of pre-commit config from main onto the chore branch, then run pre-commit autoupdate on the chore branch. This should avoid merge conflicts resulting from running autoupdate on main.

We want to run pre-commit autoupdate on the config from main, not whatever happens to already be on the chore branch.

### How Has This Been Tested?

 Test A: Create two branches locally. Commit the same file with different contents to each of them. Run `git checkout branch1 filename`. The file on the target branch should now be a copy of that on the other branch.

No

### DIFF
--- a/.github/workflows/pre_commit_autoupdate.yml
+++ b/.github/workflows/pre_commit_autoupdate.yml
@@ -58,6 +58,10 @@ jobs:
         run: |
           git checkout main
           echo "hash=$( sha256sum $config_path )" >> $GITHUB_OUTPUT
+      - name: Overwrite config on branch with version from main
+        run: |
+          git checkout ${{ env.update_branch }}
+          git checkout main ${{ env.config_path }}
       - name: Run Pre-commit autoupdate
         id: autoupdate
         run: |
@@ -71,7 +75,6 @@ jobs:
         id: commit
         if: steps.old_file.outputs.hash != steps.new_file.outputs.hash
         run: |
-          git checkout -m ${{ env.update_branch }}
           export message="chore(deps): autoupdate pre-commit hooks"
           export sha=$( git rev-parse ${{ env.update_branch }}:${{ env.config_path }} )
           export content=$( base64 -i ${{ env.config_path }} )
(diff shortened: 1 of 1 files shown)


.github/workflows/pre_commit_autoupdate.yml
```diff
@@ -58,6 +58,10 @@ jobs:
         run: |
           git checkout main
```
tc85324, line 59: Could we change this to only checkout the pre-commit config from main rather than the whole branch? This way we won't need the new step, but the desired logic will still be achieved.

tc85324 (commented):
Thank you for fixing this! It should avoid the silent merge conflict while still preserving any commits made directly to the pre-commit chore branch. I've suggested a slight modification to reduce the (albeit very few) lines of new code.

.github/workflows/pre_commit_autoupdate.yml
tp832944: Good idea! I've tweaked your suggestion slightly to ensure it's clear what's going on.

Issue #513, opened 2024-03-25
Accommodate bug fix to pre-commit hook fail_fast


tp832944:
### What's the issue?

When setting up the pre-commit config, we relied on some behaviour contrary to the documentation for hook `fail_fast`. This turned out to be a bug that has now been fixed in v3.7.0 https://github.com/pre-commit/pre-commit/issues/3167. This means that the pylint hook is now only skipped if ruff fails, rather than if any previous hook fails.

We would like to keep this behaviour because pylint is a slow hook and there is some duplication of functionality between ruff and pylint.

Thus, the action to be taken is to update the comment on `fail_fast` in the ruff hook to reflect this new behaviour. Also, update the dependencies to require at least v3.7.0 of pre-commit.

PR #526, opened 2024-03-28
fix: stop pre-commit if ruff linter fails

tp832944:
refs: #513

### PR Type

- CI related changes

### Description

Pre-commit was incorrectly terminating if ruff formatter fails; it should have been ruff linter.

No

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -103,11 +103,11 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.3.4
     hooks:
+      # Run the formatter.
+      - id: ruff-format
       # Run the linter.
       - id: ruff
         args: [ --fix ]
-      # Run the formatter.
-      - id: ruff-format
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -103,11 +103,11 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.3.4
     hooks:
+      # Run the formatter.
+      - id: ruff-format
```
tc85324, line 107: Because we run the linter with the `--fix` option, the formatting hook needs to be after the linting hook. See the [ruff-pre-commit readme](https://github.com/astral-sh/ruff-pre-commit) for more info.

tc85324 (changes requested):

.pre-commit-config.yaml
tp832944: Thanks @tc85324, I wasn't aware of that. I've reverted the order but kept `fail_fast` on the linter, which is the point of this PR.

tc85324 (approved):

PR #529, opened 2024-04-05
feat: add `Coresubset` abstract base class

### LINKED ISSUE #518: Add a class for coresubsets
### What's the new feature?

Add a `CoreSubset` class like that shown below:

```python
# Current implementation
class KernelHearding(coreax.reduction.Coreset):
    self.coreset_indices = ... # some value that is set after a call to `fit_to_size`

# Proposed implementation
class CoreSubset(coreax.reduction.Coreset):
      """Docstring explaining that a coresubset is a special type of coreset"""
      self.coreset_indices = None  # Remove from Coreset and init here as None

class KernelHearding(CoreSubset):
      self.coreset_indices = ... # some value that is overwritten from `None` after a call to `fit_to_size`
```

### What value does this add?

At the moment, a child of `Coreset` is considered as a coresubset if the `Coreset.corset_indices is not None`. This kind of implicit "typing" can be confusing for new users, and makes it impossible to determine the "type" of coreset algorithm being used at runtime (prior to fitting, all corsets have `Corset.corset_indices = None`)

*Documentation example:* if the `KernelHerding` class were not in the `coresubset.py` file, which may not be the case for future coresubset algorithms, the only way to determine that it is a coresubset (not a coreset) is to inspect the code in its `fit_to_size` method. This assumes the user is already aware that setting a `coreset_indices != None` indicates a coresubset algorithm. If `KernelHerding` was a child of a `CoreSubset` class, it is immediately obvious that it generates a coresubset, and that the important documentation regarding the difference between a `Coreset` and a `Coresubset` can be found by going to the definition of `CoreSubset`.

*Runtime example:* imagine a user wishes to construct a searchable set of `Coreset` classes/algorithms. Currently, there is no option to filter such a set to only `CoreSubsets`. However, if the coresubset methods all inherited from a `CoreSubset` class, it would be as simple as `filter(lambda x: isinstance(x, CoreSubset), searchable_set)`. You could even do something like [this](https://stackoverflow.com/questions/3862310/how-to-find-all-the-subclasses-of-a-class-given-its-name) if you wanted to programmatically identify all `CoreSubsets` implemented in coreax.

Continue with the current approach.

### Additional context

_No response_

tc85324:
### PR Type
Closes #518 
- Feature
- Tests

### Description
- Updates the `Coreset` docstring
- Moves `kernel_matrix_row_sum_mean` into `KernelHerding` (this attribute is not required for general `Coreset` algorithms)
- Adds a `Coresubset` abstract base class
- Updates existing `Coresubset` algorithms to inherit from the newly added `Coresubset` class
- Adjust previous `Coresubset` checks to test on the instance type of a given `Coresubset` (not the value of `Coresubset.corset_indices`)
- Converts `TestCoreset` to use pytest, allowing parameterization of the test on both `Coreset` and `Coresubset`
- Moves mocks used in multiple tests, `CoresetMock` and `CoresubsetMock`, into a new `tests/helpers.py`

### How Has This Been Tested?
Changes have been made to the test suite where necessary.

The implicit determination mechanisms (if `Coreset.coreset_indices is None` then is `Coresubset`) no longer work. Instead, determination of a `Coreset` as a `Coresubset` should be performed with `isinstance(coreset, Coresubset)`.

### DIFF
--- a/coreax/reduction.py
+++ b/coreax/reduction.py
@@ -55,14 +55,44 @@ from typing_extensions import Self
 import coreax.data
 import coreax.kernel
 import coreax.metrics
-import coreax.refine
 import coreax.util
 import coreax.weights
 
 
 class Coreset(ABC):
     r"""
-    Class for reducing data to a coreset.
+    Abstract base class for coresets.
+
+    TLDR: a coreset is a reduced set of :math:`\hat{n}` (potentially weighted) data
+    points that, in some sense, best represent the "important" properties of a larger
+    set of :math:`n > \hat{n}` (potentially weighted) data points.
+
+    Given a dataset :math:`X = \{x_i\}_{i=1}^n, x \in \Omega`, where each node is paired
+    with a non-negative (probability) weight :math:`w_i \in \mathbb{R} \ge 0`, there
+    exists an implied discrete (probability) measure over :math:`\Omega`
+
+    .. math:
+        \eta_n = \sum_{i=1}^{n} w_i \delta_{x_i}.
+
+    If we then specify a set of test-functions :math:`\Phi = {\phi_1, \dots, \phi_M}`,
+    where :math:`\phi_i \colon \Omega \to \mathbb{R}`, which somehow capture the
+    "important" properties of the data, then there also exists an implied push-forward
+    measure over :math:`\mathbb{R}^M`
+
+    .. math:
+        \mu_n = \sum_{i=1}^{n} w_i \delta_{\Phi(x_i)}.
+
+    A coreset is simply a reduced measure containing :math:`\hat{n} < n` updated nodes
+    :math:`\hat{x}_i` and weights :math:`\hat{w}_i`, such that the push-forward measure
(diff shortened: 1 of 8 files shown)


coreax/reduction.py
```diff
@@ -293,6 +293,110 @@ def validate_fitted(self, caller_name: str) -> None:
...
+        \hat{x}_i = x_i, \forall i \in I,
+        I \subset \{1, \dots, n\}, text{card}(I) = \hat{n}.
+
+    Thus, a coresubset, unlike a corset, ensures that feasibility constraints on the
+    support of the measure are maintained :cite:`litterer2012recombination`. This is
+    vital if, for example, the test-functions are only defined on the support of the
+    original measure/nodes, rather than all of :math:`\Omega`.
+
+    In coresubsets, the measure reduction can be implicit (setting weights/nodes to
+    zero for all :math:`i \in I \ {1, \dots, n}`) or explicit (removing entries from the
```
tc85324, line 314: Should this be ${1, \dots, n - \hat{n}}$?

Issue #420, opened 2024-02-07
[Feature]: Restructure input validation


ghost:
### Describe the feature that you would like.*

Restructure how variables are validated throughout the codebase:

- Avoid explicit validation, instead passing variables to their point of use and letting an error be raised if the value is inappropriate at the point of use.
- If one of our functions does need to validate a variable in order to guard against security risks from bad input, use explicit `if` statements in preference to hidden validation functions, e.g. `if a < 5: raise ValueError("Suitable error message")` instead of `validate_in_range`.
- Remove the module `validation.py`.

### Why is this feature valuable to you?*

- Make the code easier to read.
- Reduce processing overhead for users who supply valid input.
- Guard against abuse of private functions.

### Additional context

The codebase currently includes various validation functions in validate.py. The core development team has agreed that these should be replaced using the approach described in this ticket.

### Code of Conduct*

- [x] I agree to follow this project's Code of Conduct

tp832944:
Ticket needs to be broken down by module.

tp832944:
Closing as covered by other, more granular tickets.

tp832944:
@pc532627 Here is our strategy:

- Do cast when required, e.g. `jnp.atleast_2d`.
- Try to minimise the number of times the same thing is validated. This is easier said than done because we have many nested calls of public functions. In which case, we may just have to validate at point of use, rather than the `__init__` to classes.
- Keep `isinstance` (vanilla Python, not ours in `validation.py`) where necessary to provide useful error messages. If we're about to check for a particular attribute, it's probably not necessary to check the class is of the correct type.
- In `KernelHerding`, there's a double loop with `max_size` as the step parameter.
  - If positive, we're ok.
  - If zero, get unhelpful error.
  - If negative, fails silently with weird output as loop does not execute.
  - Define a function so that we can update `max_size` in a local scope.
  - `max_size = max(max_size, 0)`
  - `try: indices = range(from, to, max_size); except whatever error when step=0: raise ValueError("useful message")`
  - Only need to apply to one `range` as the other will yield the same error.
  - Key is a very tight `try-except` statement designed to only pick up exactly what we want to validate. Do not use an `if` statement in LBYL style if you can help it.

Issue #237, opened 2023-11-20
Test kernel.compute for different shapes of input


sk062380:
### Is your proposal related to a problem?

We should make sure kernel.compute works for different shapes of input since Stein herding uses different input shapes to the other types of herding that are implemented. 

### Describe the solution you would like*

Should just take a few unit tests to cover inputs as integer, vector, array etc.

### Code of Conduct*

- [X] I agree to follow this project's Code of Conduct

tc85324:
This was covered in #543

PR #543, opened 2024-04-10
feat: utilise pytest for `test_kernel`

### LINKED ISSUE #542: Use pytest parameterization for `test_kernel.py`
### What's the new feature?

Utilise pytest parameterization and fixtures for the repeated test cases in `test_kernel.py`

### What value does this add?

Same as for #532.

Reduces the size of the tests, better highlights where the tests for specific implementations diverge, and reduces the opportunity for mistakes in/across the tests themselves.

_No response_

### Additional context

_No response_

tc85324:
### PR Type
Closes #542

- Feature
- Tests

### Description
Significantly reduces the LOCs and enhances modularity of the tests. This should help improve test consistency and reduce the implementation burden for testing any future additions to `kernel.py`.

### How Has This Been Tested?
The new tests pass.

Removes the following redundant/ineffective tests:
- `test_approximator_valid`: test did not perform effective checks
- `test_stein_kernel_computation`: redundant overlap with `test_stein_kernel_element_computation`
- `test_stein_kernel_invalid_base_kernel_missing_divergence` and `test_stein_kernel_invalid_base_kernel_no_pytree` are replaced with a single check of `InvalidKernel`

Modifies the following logic:
- ~~`SteinKernel` now only checks on a call to `compute_elementwise` if the `base_kernel` attribute is an instance of `Kernel` (as required)~~

### DIFF
--- a/coreax/kernel.py
+++ b/coreax/kernel.py
@@ -848,23 +848,9 @@ class SteinKernel(Kernel):
         output_scale: float = 1.0,
     ):
         """Define the Stein kernel, i.e. the application of the Stein operator."""
-        # Check that the base_kernel provided has the relevant JAX pytree methods to
-        # allow functionality within this kernel
-        if not hasattr(base_kernel, "tree_flatten"):
-            raise AttributeError(
-                "base_kernel must have the method tree_flatten implemented"
-            )
-
-        if not hasattr(base_kernel, "tree_unflatten"):
-            raise AttributeError(
-                "base_kernel must have the method tree_unflatten implemented"
-            )
-
         self.base_kernel = base_kernel
         self.score_function = score_function
         self.output_scale = output_scale
-
-        # Initialise parent
         super().__init__(output_scale=output_scale)
 
     def tree_flatten(self) -> tuple[tuple, dict]:
--- a/tests/unit/test_kernel.py
+++ b/tests/unit/test_kernel.py
@@ -62,11 +62,12 @@ class _PairwiseProblemFactory(Protocol, Generic[_Kernel_contra]):
     ) -> tuple[ArrayLike, ArrayLike, Array | np.ndarray]: ...
 
 
-class _Problem(NamedTuple, Generic[_Kernel]):
+# Once we support only python 3.11+ this should be generic on _Kernel
+class _Problem(NamedTuple):
     x: ArrayLike
     y: ArrayLike
     expected_output: ArrayLike
-    kernel: _Kernel
+    kernel: Kernel
 
 
 class BaseKernelTest(ABC, Generic[_Kernel]):
@@ -81,9 +82,7 @@ class BaseKernelTest(ABC, Generic[_Kernel]):
         """
 
     @abstractmethod
-    def problem(
-        self, request, kernel_factory: _KernelFactory[_Kernel]
-    ) -> _Problem[_Kernel]:
+    def problem(self, request, kernel_factory: _KernelFactory[_Kernel]) -> _Problem:
         """Abstract pytest fixture which returns a problem for ``Kernel.compute``."""
 
     @pytest.fixture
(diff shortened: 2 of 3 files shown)


rg936672:
This has merge conflicts atm

tc85324:
Check that this covers #237

coreax/kernel.py
```diff
@@ -915,6 +901,10 @@ def compute_elementwise(
         :param y: Vector :math:`\mathbf{y} \in \mathbb{R}^d`
         :return: Kernel evaluated at (``x``, ``y``)
         """
+        if not isinstance(self.base_kernel, Kernel):
```
rg936672, line 904: As far as I'm aware, the preferred approach on Coreax at the moment is not to rely on `isinstance` checks, but rather to rely on `hasattr` checks (i.e. relying on duck typing rather than explicit typing). I'm happy for this change to go through, but you should probably get a second opinion!

tests/unit/test_kernel.py
```diff
@@ -19,1993 +19,572 @@
...
-        self.default_length_scale = 1 / np.sqrt(2)
-        self.default_x = jnp.array([0.0, 1.0, 2.0, 3.0, 4.0]).reshape(-1, 1)
-        self.default_zero_kernel_row_sum = jnp.zeros(len(self.default_x))
-
-        # Define the simplest, real kernel object for testing purposes
-        self.default_kernel = coreax.kernel.SquaredExponentialKernel(
-            length_scale=self.default_length_scale
-        )
-
-    def test_approximator_valid(self) -> None:
```
rg936672, line 194: We seem to have lost this test!

rg936672 (changes requested):
As well as the couple of comments I left, the main issue is that we've lost most of the test docstrings in the transition to pytest - if these are added back (as well as the missing test) this should be good to merge. This is nonetheless an impressive shortening!

tests/unit/test_kernel.py
tc85324: I can add this back in, but it will be deleted again if #583 is merged first as the functionality being tested will no longer exists (in this form). I had also been sceptical of the utility of this test, as it is only checking that the mocked approximator is being called, not that approximation is working (although I guess this is the only way to to have the test success decoupled from the success of the approximator implementation)

coreax/kernel.py
tc85324: What do you think @tp832944 ?

tp832944: The agreed method (see https://github.com/gchq/coreax/issues/420#issuecomment-1961033259) is to just let it crash when safe to do so. Otherwise, use `try ... except`. If the error message from letting it crash is unhelpful, use `try ... except` then reraise a useful error. In this case, the attribute error raised by the next line (908) is fairly informative, so just delete 904-907.

rg936672: On the topic of helpful vs unhelpful error messages, I seem to remember that trying to JIT an object missing `tree_flatten` or `tree_unflatten` gave a really unhelpful error message. However, if we're moving to `equinox` this shouldn't be an issue.

tc85324: I'll remove the validation change from this pull and we can have any other discussions regarding validation in #583

tc85324:
I've removed the validation, added back in the docstrings (where appropriate), and have enhanced the accuracy (and hopefully clarity) of the typehints.

rg936672:
Oops. test errors on 3.9 and 3.10 (`TypeError: Multiple inheritance with NamedTuple is not supported`)

rg936672 (approved):
Once the test collection failures are fixed (should be easily resolved by making `_Problem` not inherit from `NamedTuple`), this looks good to go! 

In future, though, I'd ask that you don't amend or rebase commits that have already been reviewed if possible - it makes reviewing incremental changes (i.e. changes only since the last commit reviewed) more of a hassle.

tc85324:
Sorry about that, I had realised it would be a problem straight after pushing. I've removed the generic from the `_Problem` tuple to fix collection on versions less than 3.11.

rg936672 (approved):

PR #547, opened 2024-04-11
feat: utilise pytest for `test_refine`

### LINKED ISSUE #546: Use pytest parameterization for `test_refine.py`
### What's the new feature?

Utilise pytest parameterization and fixtures for the repeated test cases in `test_refin.py`

### What value does this add?

Same as for #532 and #542

Reduces the size of the tests, better highlights where the tests for specific implementations diverge, and reduces the opportunity for mistakes in/across the tests themselves.

_No response_

### Additional context

_No response_

tc85324:
### PR Type
Closes #546

- Feature
- Tests

### Description
Reduces the LOC and enhances the modularity of the tests. This should help improve test consistency and reduce the implementation burden for testing any future additions to refine.py.

### How Has This Been Tested?
The new tests pass.

N/A

### DIFF
--- a/tests/unit/test_refine.py
+++ b/tests/unit/test_refine.py
@@ -20,12 +20,16 @@ within this file verify that refinement approaches used produce the expected res
 simple examples.
 """
 
+from __future__ import annotations
+
 import itertools
-import unittest
+from abc import ABC, abstractmethod
 from unittest.mock import patch
 
 import jax.numpy as jnp
-from jax import random
+import jax.random as jr
+import pytest
+from jax import Array
 
 import coreax.approximation
 import coreax.data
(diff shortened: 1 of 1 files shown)


tests/unit/test_refine.py
```diff
@@ -38,237 +42,56 @@
...
-    """
-    Tests related to :meth:`~coreax.refine.RefineRegular`.
-    """
-
-    def setUp(self):
-        self.random_key = random.key(0)
-
-    def test_refine_ones(self) -> None:
-        """
-        Test that refining an optimal coreset leaves ``coreset_indices`` unchanged.
```
rg936672, line 101: I don't think we're actually testing this any more?

tests/unit/test_refine.py
```diff
@@ -280,160 +103,92 @@ def test_valid_coreset_no_kernel(self):
...
+            (
+                jnp.asarray([[0, 0], [1, 1], [0, 0], [1, 1]]),
+                list(itertools.combinations(range(4), 2)),
+                [{0, 1}, {0, 3}, {1, 2}, {2, 3}],
+            ),
+            (jnp.asarray([[0, 0], [1, 1], [2, 2]]), [[2, 2]], [{0, 2}]),
+        ],
+    )
+    @pytest.mark.parametrize("use_cached_row_sum_mean", [False, True])
+    def test_refine(
```
rg936672, line 130: The parametrisation of this test is hard to parse - it needs more explanation attached, and might be better off split into two separate tests (as the tests they come from are testing subtly different things).

rg936672 (changes requested):
Overall this looks good, but I think one of the tests has been somewhat lost in the transition

tests/unit/test_refine.py
```diff
@@ -452,43 +207,15 @@ def test_zero_original_data_points(self):
...
-        # raise a value error highlighting the root cause
-        refine_random = coreax.refine.RefineRandom(self.random_key, p=-0.5)
-        with self.assertRaises(ValueError) as error_raised:
-            refine_random.refine(coreset=coreset_obj)
-
-        self.assertEqual(
-            error_raised.exception.args[0],
-            "input p must be greater than 0",
-        )
+        Test how RefineRandom on different proportions of the original data to sample.
```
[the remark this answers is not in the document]
tc85324: Good point!

tests/unit/test_refine.py
tc85324: So I wasn't sure what this test was trying to check. As far as I'm aware, we don't expect refine to have the machinery to shortcut the expensive parts of the refinement step if `coreset_indices = best_indices`. I.E. we still expect refinement to proceed as usual even if we started with the "correct" coreset indices. In this test we have multiple feasible values for the `best_indices`, `{0, 1}` or `{0, 3}` or `{1, 2}` or `{2,3}`, so are we expecting refine to enumerate all of the possible `best_indices` and only return the original coreset indices if they were one of the possible `best_indices`?

I.E. I don't think this test is really checking anything different to `test_refine_ints`.

rg936672: The difference is that in this test we start with an already-optimal coreset, and we're checking that it's not changed (i.e. to a different optimal coreset). If we guarantee this stability/idempotency property, then it makes sense to have this test in place; if we don't, the current implementation as a parametrisation of test_refine (which doesn't check that the coreset_indices are unchanged) is OK

tests/unit/test_refine.py
tc85324: Looking at this again, I totaly agree. This is confusing, especially as it wasn't clear that this addressed the `test_refine_ones`. I'll have a think about a nicer way to do this.

tests/unit/test_refine.py
tc85324: I can not identify any functionality within the existing refine methods that would guarantee the idempotency property, nor can I find any documentation suggesting that this is a required/expected behaviour (excluding the old test), so I believe the current parameterisation of `test_refine` should suffice.

rg936672 (approved):

PR #563, opened 2024-04-18
fix: handle very rare `jit_test` test failures

### LINKED ISSUE #525: Random failure of test_util.TestUtil.test_jit_test
### What's the problem?

Line 200: `self.assertLess(post_time, (pre_time - wait_time))` has a comment above saying that this is almost certain to hold. This assertion may break if there is a scheduling delay on the processor during the second run to calculate `post_time`.

Although still not guaranteed to pass, a time delta should be added, e.g.  `self.assertLess(post_time, (pre_time - 0.5 * wait_time))`.

We may also want to add such a margin to `coreax.util.jit_test` or into the performance tests.

Failures are rare. An example is https://github.com/gchq/coreax/actions/runs/8455302667/job/23162458861.

### Python version

3.9, 3.10, 3.11

### Package version

0.1.0

### Operating system

All

### Other packages

_No response_

### Relevant log output

_No response_

tc85324:
### PR Type
Closes #525

- Bugfix
- Tests

### Description
Handles the very rare failures of the `jit_test` test by utilising `pytest.mark.flaky` from `pytest-rerunfailures`, which re-runs the test on the event of a failure (up to three times).

Now uses pytest parameterisation such that each individual scenario can be re-run if failed, rather than re-running until every scenario passes at once (reducing the probability of continued failure after three re-runs).

Additionally, provides an explicit check to ensure the function is only traced once (the JIT compiled function is being called on the second call) to supplement the implicit checks performed in checked the timings.

### How Has This Been Tested?
The new/updated unit tests pass.

No.

### DIFF
--- a/.cspell/library_terms.txt
+++ b/.cspell/library_terms.txt
@@ -85,6 +85,7 @@ pytrees
 quickstart
 rcfile
 rect
+rerunfailures
 rightarrow
 rtol
 samp
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,6 +44,7 @@ test = [
     "opencv-python",
     "pyqt5",
     "pytest-cov",
+    "pytest-rerunfailures",
     "scipy",
 ]
 # Compile documentation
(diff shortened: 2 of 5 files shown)


pc532627 (approved):
Changes look good, happy to merge.

PR #579, opened 2024-04-25
feat: added linear kernel with associated tests

db091756:
### PR Type

- Feature
- Tests

### Description

Adds the Linear kernel.

### How Has This Been Tested?

Unit tests for the linear kernel in the same style as the currently implemented kernels.

No

### Screenshots

NA

### DIFF
--- a/.cspell/custom_misc.txt
+++ b/.cspell/custom_misc.txt
@@ -19,7 +19,6 @@ kdtree
 kernelised
 kernelized
 ml.p3.8xlarge
-parameterizable
 parsable
 PCIMQ
 PMLR
--- a/.cspell/custom_misc.txt
+++ b/.cspell/custom_misc.txt
@@ -19,7 +19,7 @@ kdtree
 kernelised
 kernelized
 ml.p3.8xlarge
-parameterisable
+parameterizable
 parsable
 PCIMQ
 PMLR
--- a/.cspell/custom_misc.txt
+++ b/.cspell/custom_misc.txt
@@ -19,6 +19,7 @@ kdtree
 kernelised
 kernelized
 ml.p3.8xlarge
+parameterizable
 parsable
 PCIMQ
 PMLR
--- a/.cspell/custom_misc.txt
+++ b/.cspell/custom_misc.txt
@@ -19,6 +19,7 @@ kdtree
 kernelised
 kernelized
 ml.p3.8xlarge
+parameterisable
 parsable
 PCIMQ
 PMLR
@@ -30,4 +31,3 @@ RKHS
 RPCHOLESKY
 sigmas
 WMMd
-parameterisable
--- a/.cspell/custom_misc.txt
+++ b/.cspell/custom_misc.txt
@@ -30,3 +30,4 @@ RKHS
 RPCHOLESKY
 sigmas
 WMMd
+parameterisable
--- a/coreax/kernel.py
+++ b/coreax/kernel.py
@@ -24,7 +24,7 @@ In this library, we often use kernels as a smoothing tool: given a dataset of di
 points, we can reconstruct the underlying data generating distribution through smoothing
 of the data with kernels.
 
-Some kernels are parameterizable and may represent other well known kernels when given
+Some kernels are parameterisable and may represent other well known kernels when given
 appropriate parameter values. For example, the :class:`SquaredExponentialKernel`,
 
 .. math::
@@ -304,6 +304,47 @@ class Kernel(eqx.Module):
             block_row_sum = jnp.r_[block_row_sum.reshape(-1), trailing_block_row_sum]
         row_sum = jnp.sum(block_row_sum.reshape(-1, num_data_points), axis=0)
         return row_sum / num_data_points
+
+
+class LinearKernel(Kernel):
+    """
+    Define a linear kernel.
+
+    :param length_scale: Kernel smoothing/bandwidth parameter
+    :param output_scale: Kernel normalisation constant
+    """
+
+    @override
+    def compute_elementwise(
+        self,
+        x: ArrayLike,
+        y: ArrayLike,
+    ) -> Array:
+        return x.dot(y)
+
+    @override
+    def grad_x_elementwise(
+        self,
+        x: ArrayLike,
+        y: ArrayLike,
+    ) -> Array:
+        return y
+
+    @override
+    def grad_y_elementwise(
+        self,
+        x: ArrayLike,
+        y: ArrayLike,
+    ) -> Array:
+        return x
+
+    @override
+    def divergence_x_grad_y_elementwise(
+        self,
+        x: ArrayLike,
+        y: ArrayLike,
+    ) -> Array:
+        return x.shape[0]
 
 
 class SquaredExponentialKernel(Kernel):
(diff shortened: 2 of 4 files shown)


coreax/kernel.py
```diff
@@ -24,7 +24,7 @@
 points, we can reconstruct the underlying data generating distribution through smoothing
 of the data with kernels.
 
-Some kernels are parameterizable and may represent other well known kernels when given
```
tc85324, line 27: The old spelling is correct British English as defined by the OED.

coreax/kernel.py
```diff
@@ -348,6 +348,45 @@ def calculate_kernel_matrix_row_sum_mean(
...
 
 
+class LinearKernel(Kernel):
+    """Define a linear kernel."""
+
+    length_scale: float = 1.0
+    output_scale: float = 1.0
+
+    @override
+    def compute_elementwise(
```
tc85324, line 358: Remove the comma after each `y: Arraylike`, then when you re-run ruff format these should all fit on one line (see the base class for example).

coreax/kernel.py
```diff
@@ -348,6 +348,45 @@ def calculate_kernel_matrix_row_sum_mean(
         return self.calculate_kernel_matrix_row_sum(x, max_size) / (1.0 * x.shape[0])
 
 
+class LinearKernel(Kernel):
+    """Define a linear kernel."""
```
tc85324, line 352: I know the docstrings for the other kernels are equally sparse, but is there any chance you could add the mathematical definition in here as well? (perhaps also some general comments on the utility of this kernel) It would then be good if you could add these same definitions to the other kernels (in another pull).

tests/unit/test_kernel.py
```diff
@@ -284,6 +285,99 @@ def expected_divergence_x_grad_y(
...
+    @pytest.fixture
+    def kernel_factory(self) -> _KernelFactory[LinearKernel]:
+        def kernel(length_scale, output_scale):
+            return LinearKernel(length_scale, output_scale)
+
+        return kernel
+
+    @override
+    @pytest.fixture(
+        params=[
```
tc85324, line 304: Please remove the trailing comma from arrays. Ruff will then format this on one line.

coreax/kernel.py
```diff
@@ -348,6 +348,45 @@ def calculate_kernel_matrix_row_sum_mean(
...
+    length_scale: float = 1.0
+    output_scale: float = 1.0
+
+    @override
+    def compute_elementwise(
+        self,
+        x: ArrayLike,
+        y: ArrayLike,
+    ) -> Array:
+        return x.dot(y)
```
tc85324, line 363: Pyright rightly raises and errpr here because the `ArrayLike` type hint is very permissive. For example, if x is an integer, it won't have a `x.dot` attribute. Instead you can use  `jnp.dot(x, y)`.

coreax/kernel.py
```diff
@@ -348,6 +348,45 @@ def calculate_kernel_matrix_row_sum_mean(
...
+
+    @override
+    def grad_x_elementwise(
+        self,
+        x: ArrayLike,
+        y: ArrayLike,
+    ) -> Array:
+        return y
```
tc85324, line 371: Similar to above. To actually be returning an `Array`, y needs to be cast using `jnp.asarray(y)`. Same for grad_y_elementwise.

coreax/kernel.py
```diff
@@ -348,6 +348,45 @@ def calculate_kernel_matrix_row_sum_mean(
...
+    ) -> Array:
+        return x
+
+    @override
+    def divergence_x_grad_y_elementwise(
+        self,
+        x: ArrayLike,
+        y: ArrayLike,
+    ) -> Array:
+        return x.shape[0]
```
tc85324, line 387: Again, if `x` were an int/float, it wouldn't have a shape attribute. Instead you should use `jnp.asarray(jnp.shape(x)[0])` (the as array is needed to convert the int to an array, as required by the type hint)

documentation/source/index.rst
```diff
@@ -2,7 +2,7 @@ Coreax (|version|)
 ==================
 
 Coreax is a library for **coreset algorithms**, written in
-:doc:`JAX <jax:notebooks/quickstart>` for fast execution and GPU support.
+:doc:`JAX <jax:quickstart>` for fast execution and GPU support.
```
tc85324, line 5: This change shouldn't be in this pull.

tc85324 (changes requested):
Generally looks good! Some minor comments about ensuring type hints are properly followed and that documentation is sufficiently descriptive.

coreax/kernel.py
db091756: Weird, fixed

coreax/kernel.py
db091756: Whoops, fixed

tests/unit/test_kernel.py
db091756: Fixed

coreax/kernel.py
db091756: Fixed

coreax/kernel.py
db091756: Added definition in there but left the utility comments off because I'm not aware of any advice around when linear kernel should be used when building coresets vs other kernels.

coreax/kernel.py
```diff
@@ -348,6 +348,34 @@ def calculate_kernel_matrix_row_sum_mean(
...
+    def grad_x_elementwise(self, x: ArrayLike, y: ArrayLike) -> Array:
+        return jnp.asarray(y)
+
+    @override
+    def grad_y_elementwise(self, x: ArrayLike, y: ArrayLike) -> Array:
+        return jnp.asarray(x)
+
+    @override
+    def divergence_x_grad_y_elementwise(self, x: ArrayLike, y: ArrayLike) -> Array:
+        return jnp.asarray(x).shape[0]
```
tc85324, line 376: I like this better than my suggestion!

.cspell/custom_misc.txt
```diff
@@ -18,6 +18,7 @@ kdtree
 kernelised
 kernelized
 ml.p3.8xlarge
+parameterizable
```
tc85324, line 21: This is now redundant.

coreax/kernel.py
```diff
@@ -348,6 +348,34 @@ def calculate_kernel_matrix_row_sum_mean(
         return self.calculate_kernel_matrix_row_sum(x, max_size) / (1.0 * x.shape[0])
 
 
+class LinearKernel(Kernel):
+    r"""
+    Define a linear kernel.
+
+    The linear kernel is defined as :math:`k: \mathbb{R}^d\times \mathbb{R}^d
+    \xrightarrow{} \mathbb{R}`, :math:`k(x, y) = x^Ty`.
```
tc85324, line 356: Can we use `\to` instead of `\xrightarrow{}`?

tc85324 (changes requested):
Very minor complaints. Otherwise, looks good!

coreax/kernel.py
tc85324: Please can you open an issue to add these descriptions to the other kernel docstrings.

tc85324 (approved):
Looks good!

PR #601, opened 2024-05-07
Refactored metrics.py to use new data classes

db091756:
### PR Type

- Feature
### Description

Refactors metrics.py to make use of new data classes.
Small change to `weighted_maximum_mean_discrepancy` to make the function symmetric.

### How Has This Been Tested?

unit tests also refactored

(Write your answer here.)

### Screenshots

(Write your answer here.)


rg936672 (dismissed):
The tests are currently all failing with `TypeError: Cannot interpret value of type <class 'memoryview'> as an abstract array; it does not have a dtype attribute` - this needs to be fixed before we can merge! I'll have a look at the rest of the PR in the meantime, though.

tests/unit/test_coresubset.py
```diff
@@ -116,6 +116,7 @@ def test_fit_comparison_to_random_and_refined(self) -> None:
         # to typically compete with a kernel herding approach
         generator = np.random.default_rng(self.random_data_generation_key)
         x = generator.random((num_data_points, self.dimension))
+        metric_x = coreax.data.Data(x)
```
db091756, line 119: Bit hacky but can be cleaned up when reduction is changed to use the new data classes

coreax/metrics.py
```diff
@@ -95,86 +91,104 @@ def __init__(self, kernel: coreax.kernel.Kernel, precision_threshold: float = 1e
...
+            - 2\mathbb{E}(k(\mathcal_{D}_1,\mathcal_{D}_2))
 
         where :math:`k` is the selected kernel.
 
-        :param x: The original :math:`n \times d` data
-        :param y: An :math:`m \times d` array defining a representation of ``x``, for
-            example a coreset
+        :param reference_points: The original :math:`n \times d` data
+        :param comparison_points: An :math:`m \times d` array to compare to
+            ``reference_points``
```
rg936672, line 181: We've lost the extra context provided by the original docstring - is this intentional? Same question for the other MMD methods

coreax/metrics.py
```diff
@@ -95,86 +91,104 @@ def __init__(self, kernel: coreax.kernel.Kernel, precision_threshold: float = 1e
...
+                )
+            return self.maximum_mean_discrepancy_block(
+                reference_points, comparison_points, block_size
+            )
 
         if (
             block_size is None
-            or weights_x is None
-            or block_size > max(num_points_x, num_points_y)
+            or reference_weights is None
```
rg936672, line 146: I recognise this is a pre-existing issue, but if `block_size is None` and `reference_weights is not None` then `reference_weights` is not used. If this is intended behaviour, it should at least warn the user!

coreax/metrics.py
```diff
@@ -95,86 +91,104 @@ def __init__(self, kernel: coreax.kernel.Kernel, precision_threshold: float = 1e
...
 
-        If no weights are given for dataset ``y``, standard MMD is calculated. If
-        weights are given, weighted MMD is calculated. For both cases, if the size of
-        matrix blocks to process is less than the size of both datasets and if
-        ``weights_x`` is given in the weighted case, the calculation
-        is done block-wise to limit memory requirements.
+        If uniform weights are given for dataset ``reference_data``, standard MMD is
+        calculated. If non-uniform weights are given, weighted MMD is calculated. For
+        both cases, if the size of matrix blocks to process is less than the size of
+        both datasets and reference weights are not uniform in the weighted case, the
```
rg936672, line 104: I think this could be an issue in future - at present it's not possible to do the calculation block-wise with uniform weights, whereas before this change there was a difference between passing in _unweighted_ data (which could not be done block-wise) and _uniformly-weighted_ data (which could, as `weights_x is not None`).

tests/unit/test_coresubset.py
rg936672: Should add a ticket for the reduction refactoring, and mention on it that this change needs to made

tests/unit/test_metrics.py
```diff
@@ -323,29 +385,34 @@ def test_weighted_mmd_rand(self) -> None:
...
+            comparison_weights=self.comparison_weights,
+        )
 
         # Check output matches expected
         self.assertAlmostEqual(output, expected_output, places=5)
 
     def test_weighted_mmd_uniform_weights(self) -> None:
         r"""
-        Test that weighted MMD equals MMD if weights are uniform, :math:`w_y = 1/m`.
+        Test that weighted MMD equals MMD if weights are uniform.
```
rg936672, line 415: why this removal?

tests/unit/test_metrics.py
```diff
@@ -978,13 +1128,19 @@ def test_compute_float_block_size(self) -> None:
...
+                block_size=2.0,
+            )
         self.assertEqual(
             error_raised.exception.args[0],
             "block_size must be a positive integer",
         )
 
 
+# pylint: enable=duplicate-code
+# pylint: enable=too-many-locals
```
rg936672, line 1143: These two seem redundant - they don't have a corresponding `disable` in this file

tests/unit/test_reduction.py
```diff
@@ -143,15 +143,9 @@ def test_compute_metric(self):
...
+        coreset.coreset = jnp.array([[1], [2]])
         coreset.compute_metric(metric, block_size)
-        metric.compute.assert_called_once_with(
-            coreset.original_data.pre_coreset_array,
-            coreset.coreset,
-            block_size=block_size,
-            weights_x=None,
-            weights_y=None,
-        )
+        metric.compute.assert_called_once()
```
rg936672, line 148: This test doesn't really check much of anything - are we able to go back to using `assert_called_once_with` instead of just `assert_called_once`?

rg936672 (commented):
A couple of minor docstring issues, and some issues that may be outside the scope of this PR. If they are out of scope, then create the relevant tickets before merging. Once the minor issues are fixed, and the more substantial ones are either fixed or ticketed, I'm happy for this to be merged.

coreax/metrics.py
```diff
@@ -26,57 +26,53 @@
...
-            example a coreset
-        :param block_size: Size of matrix block to process, or :data:`None` to not split
-            into blocks
-        :param weights_x: An :math:`1 \times n` array of weights for associated points
-            in ``x``, or :data:`None` if not required
-        :param weights_y: An :math:`1 \times m` array of weights for associated points
-            in ``y``, or :data:`None` if not required
+        :param reference_data: An instance of the class :class:`coreax.data.Data`
+        :param comparison_data: An instance of the class :class:`coreax.data.Data` to
+            compare against ``reference_data``
```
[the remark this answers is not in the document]
db091756: This base class can now take in instances of Data or SupervisedData here and so it is not obvious what the dataset will actually entail, i.e. we don't know if we will have a supervision array here or not. I could explicitly mention and point towards the docs for those classes as the dimensions are mentioned there? Or open to suggestions

coreax/metrics.py
db091756: Could you write that ticket @tc85324 as you have more knowledge about the row_mean implementation?

coreax/metrics.py
```diff
@@ -95,86 +91,104 @@ def __init__(self, kernel: coreax.kernel.Kernel, precision_threshold: float = 1e
...
+        calculated. If non-uniform weights are given, weighted MMD is calculated. For
+        both cases, if the size of matrix blocks to process is less than the size of
+        both datasets and reference weights are not uniform in the weighted case, the
+        calculation is done block-wise to limit memory requirements.
 
-        :param x: The original :math:`n \times d` data
-        :param y: :math:`m \times d` dataset
+        :param reference_data: An instance of the class :class:`coreax.data.Data`
+        :param comparison_data: An instance of the class :class:`coreax.data.Data` to
+            compare against ``reference_data``
```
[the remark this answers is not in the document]
db091756: Similar to above, could point towards the class docs as they have dimensions for the data array and the weights array?

coreax/metrics.py
db091756: Yes this was intentional as there is no need for comparison_points to be a coreset of reference_points. A valid use case of something like Kernel Herding would be to build coresets of two separate datasets and then compute MMD between these two coresets to do a faster test of dataset similarity; here the docstring wouldn't make sense. Moreover, the functions are symmetric in their _points arguments anyway i.e. even if we had a coreset, the coreset could go first and nothing would be different.

tests/unit/test_metrics.py
db091756: Agree, fixed

coreax/metrics.py
db091756: Yes this seems wrong, Tom C is working a sweeping change for this

tests/unit/test_coresubset.py
db091756: Added basic ticket for this #612

tests/unit/test_metrics.py
db091756: Fixed this!

coreax/metrics.py
rg936672: Fair enough! I'll point out that `weighted_maximum_mean_discrepancy` is not symmetric, but that should be obvious from the signature. On further inspection, the "...for example a coreset" comment also appears in the docstrings of `maximum_mean_discrepancy_block` and `weighted_maximum_mean_discrepancy_block` - do we want to remove these as well? If so, I think it could be sensible to include a quick line in the class docstring or in the docstring for `compute` explaining possible usages (coreset vs. whole dataset, coreset vs. coreset etc)

coreax/metrics.py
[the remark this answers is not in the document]
rg936672: Possibly yes - I think that it's important to convey here that the datasets don't have to be the same size (in terms of number of points) but they _do_ have to be the same dimension. If that's not true, then say so in the docstring, but also highlight that some metrics (i.e. MMD) will require that they are the same dimension. 

Unsure about SupervisedData - at present our only metric, MMD, ignores the supervision data, so it shouldn't make a difference whether we pass Data or SupervisedData.

coreax/metrics.py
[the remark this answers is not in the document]
rg936672: Could do, but I do think having the expected dimensions in the function docstring is good for immediate ease-of-use to an end user (especially in an IDE). Also see my response to the similar comment above, but since this is a concrete implementation rather than an abstract base class, we don't have to worry about other `Metric` implementations possibly having different requirements.

coreax/metrics.py
db091756: Yeah the change that Tom C is working on should change a lot of this so probably worth leaving until thats been pushed

coreax/metrics.py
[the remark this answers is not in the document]
db091756: Ah I see what you mean, I will note this explicitly

coreax/metrics.py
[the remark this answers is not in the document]
db091756: Agreed, added

coreax/metrics.py
db091756: I have changed this behaviour slightly but will be addressed more fully by #613

coreax/metrics.py
rg936672: I think this issue is now resolved; we should correctly call `maximum_mean_discrepancy_block` if uniform weights are passed in along with a `block_size`. However, the docstring now needs to be updated to account for this.

tc85324 (approved):
Looks good, thanks!

Issue #581, opened 2024-04-26
Refactor reduction strategy code to enable JIT of entire coreset construction


pc532627:
### What's the new feature?

Related to issues #534 & #544 - update the implementations of reduction strategies to allow the entire codebase to be JIT'ed using Equinox, which enables fast computation of several coresets. 

Clean up test_coresubset.py to not have to define a metric_x 

### What value does this add?

Improved integration with standard JAX approaches & removes side-effects that interfere with expected JIT performance.

PR #607, opened 2024-05-09
feat: add coreset and coresubset data structures

### LINKED ISSUE #518: Add a class for coresubsets

tc85324:
### PR Type
Closes #518
- Feature
- Tests

### Description
As part of #581, it is necessary to decouple the coreset data-structures and solvers/reduction-strategies. The new `coreax.coreset.Coreset` and `coreax.coreset.Coresubset` data-structures introduced here facilitate this decoupling, such that the conceptual notion of a coreset better reflects our implementation. I.E. a coreset is a data-structure and a solver is a solver, the current implementation treats a coreset as both of these things.

Note: these new data-structures are currently unused. They will be used in an upcoming PR to fully address #581.

A new set of unit tests have been created for these new data-structures in `tests/unit/test_coreset.py`.

Additionally updates the new `coreax.data.Data` with `at_least{1,2}d` equinox converters and a `__len__` magic method. This facilitates the coreset `__len__` magic method, and will allow many `atleast_{1,2}d` calls to be removed from the codebase, at a later date.

Tests for `coreax.data.Data` have been refactored to reduce code duplication resulting from a new test for the new `__len__` method.

### How Has This Been Tested?
All tests pass

No

### DIFF
--- a/coreax/coreset.py
+++ b/coreax/coreset.py
@@ -2,11 +2,15 @@
 
 from __future__ import annotations
 
+from typing import Generic, TypeVar
+
 import equinox as eqx
 import jax.numpy as jnp
 from jaxtyping import Array, Shaped
 
 from coreax.data import Data
+
+_Data = TypeVar("_Data", bound=Data)
 
 
 def _convert_to_weighted(data: Data | Array):
@@ -15,7 +19,7 @@ def _convert_to_weighted(data: Data | Array):
     return Data(data)
 
 
-class Coreset(eqx.Module):
+class Coreset(eqx.Module, Generic[_Data]):
     r"""
     Data structure for representing a coreset.
 
@@ -58,7 +62,7 @@ class Coreset(eqx.Module):
     """
 
     nodes: Data = eqx.field(converter=_convert_to_weighted)
-    pre_coreset_data: Data
+    pre_coreset_data: _Data
 
     def __len__(self):
         """Return Coreset size/length."""
@@ -70,7 +74,7 @@ class Coreset(eqx.Module):
         return self.nodes
 
 
-class Coresubset(Coreset):
+class Coresubset(Coreset[_Data], Generic[_Data]):
     r"""
     Data structure for representing a coresubset.
 
--- a/coreax/coreset.py
+++ b/coreax/coreset.py
@@ -1,0 +1,114 @@
+"""Module for defining coreset data structures."""
+
+from __future__ import annotations
+
+import equinox as eqx
+import jax.numpy as jnp
+from jaxtyping import Array, Shaped
+
+from coreax.data import Data
+
+
+def _convert_to_weighted(data: Data | Array):
+    if isinstance(data, Data):
+        return data
+    return Data(data)
+
+
+class Coreset(eqx.Module):
+    r"""
+    Data structure for representing a coreset.
+
+    TLDR: a coreset is a reduced set of :math:`\hat{n}` (potentially weighted) data
+    points that, in some sense, best represent the "important" properties of a larger
+    set of :math:`n > \hat{n}` (potentially weighted) data points.
+
+    Given a dataset :math:`X = \{x_i\}_{i=1}^n, x \in \Omega`, where each node is paired
+    with a non-negative (probability) weight :math:`w_i \in \mathbb{R} \ge 0`, there
+    exists an implied discrete (probability) measure over :math:`\Omega`
+
+    .. math:
+        \eta_n = \sum_{i=1}^{n} w_i \delta_{x_i}.
+
+    If we then specify a set of test-functions :math:`\Phi = {\phi_1, \dots, \phi_M}`,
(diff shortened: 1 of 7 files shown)


coreax/coreset.py
```diff
@@ -0,0 +1,114 @@
...
+    return Data(data)
+
+
+class Coreset(eqx.Module):
+    r"""
+    Data structure for representing a coreset.
+
+    TLDR: a coreset is a reduced set of :math:`\hat{n}` (potentially weighted) data
+    points that, in some sense, best represent the "important" properties of a larger
+    set of :math:`n > \hat{n}` (potentially weighted) data points.
```
db091756, line 24: Could stick in a >>> here to signify that the coreset is supposed to be much smaller than n

db091756 (changes requested):
All looks good to me, though I think we need a SupervisedCoreset and SupervisedCoresubset class to handle supervised coresets with the SupervisedData class

coreax/coreset.py
tc85324: It depends if you would give a set where `n` is with `>` but not `>>` `\hat{n}` a different name? If not, then I think the current notation is more accuracte.

tc85324:
If we want the ability to distinguish between supervised and unsupervised coresets, then we can make the `pre_coreset_data` generic on `_Data = TypeVar("_Data", bound=Data)`. Then you could type `Coreset[Data]` or `Coreset[SupervisedData]`. The materialized coreset is independent of the supervision, so we only need it for typing `pre_coreset_data`.

coreax/coreset.py
db091756: I guess philosophically if you've got n=1,000,000 and hat(n)=999,999 then its not much of a coreset, but i take your point and happy to leave as is!

coreax/coreset.py
```diff
@@ -0,0 +1,118 @@
...
+        coresubset nodes should be accessed via :meth:`Coresubset.coreset`.
+    :param pre_coreset_data: The dataset :math:`X` used to construct the coreset.
+    """
+
+    # Incompatibility between Pylint and eqx.field. Pyright handles this correctly.
+    # pylint: disable=no-member
+    @property
+    def coreset(self) -> Data:
+        """Materialise the coresubset from the indices and original data."""
+        coreset_nodes = self.pre_coreset_data.data[self.unweighted_indices]
```
db091756, line 110: Maybe I am misunderstanding, but if this is a supervised Coresubset then we won't be materialising the supervision here right?

tc85324: This was by design. My logic was that the coreset should only be the weighted nodes (this is consistent with the docstrings). If you require the supervision associated with each coreset node/index, then you can get this with `Coreset.pre_coreset_data.supervision[self.unweighted_indices]`, providing the `pre_coreset_data` is `SupervisedData`.

db091756: I see, makes sense!

db091756 (approved):

PR #614, opened 2024-05-10
Kernel compositions

db091756:
### PR Type

- Feature
- Tests

### Description

Adds capability to arbitrarily add, multiply and take powers of kernels, including the corresponding magic methods. Division and subtraction not included as positive semi-definiteness is not conserved with these operations.

### How Has This Been Tested?

All additional unit tests pass.

(Write your answer here.)

### Screenshots

(Write your answer here.)


tests/unit/test_kernel.py
```diff
@@ -202,22 +208,375 @@ def expected_divergence_x_grad_y(
...
+    @override
+    @pytest.fixture(scope="class")
+    def kernel(self) -> _MultiCompositeKernel:
+        """Define an ProductKernel composed of mocked kernel functions."""
+        # Variable rename allows for nicer automatic formatting
+        num_points, dimension = self.mock_num_points, self.mock_dimension
+
+        k1 = MagicMock(spec=Kernel)
+        k1.compute_elementwise.return_value = jnp.array(1.0)
+        k1.compute.return_value = 1.0 * jnp.ones((num_points, num_points))
```
tc85324, line 406: Why do all these multiply `jnp.ones` by `1.0`?

tests/unit/test_kernel.py
```diff
@@ -202,22 +208,375 @@ def expected_divergence_x_grad_y(
         """Compute expected divergence of the kernel w.r.t ``x`` gradient ``y``."""
 
 
+class KernelCompositionTest(ABC, Generic[_MultiCompositeKernel]):
```
tc85324, line 211: Does this need to inherit from `ABC`?

tests/unit/test_kernel.py
```diff
@@ -202,22 +208,375 @@ def expected_divergence_x_grad_y(
...
+        k1.grad_x_elementwise.return_value = 1.0 * jnp.ones(dimension)
+        k1.grad_x.return_value = 1.0 * jnp.ones((num_points, num_points, dimension))
+        k1.grad_y_elementwise.return_value = 1.0 * jnp.ones(dimension)
+        k1.grad_y.return_value = 1.0 * jnp.ones((num_points, num_points, dimension))
+        k1.divergence_x_grad_y_elementwise.return_value = jnp.array(1.0)
+        k1.divergence_x_grad_y.return_value = 1.0 * jnp.ones((num_points, num_points))
+
+        k2 = MagicMock(spec=Kernel)
+        k2.compute_elementwise.return_value = jnp.array(2.0)
+        k2.compute.return_value = 2.0 * jnp.ones((num_points, num_points))
```
tc85324, line 255: Instead of multiplying by 2.0 we can use `jnp.full` instead here

coreax/kernel.py
```diff
@@ -102,6 +102,55 @@ def median_heuristic(x: ArrayLike) -> Array:
 class Kernel(eqx.Module):
     """Abstract base class for kernels."""
 
+    def __add__(self, addition: Kernel | int | float):
+        """Overload `+` operator."""
+        if not isinstance(addition, (Kernel, int, float)):
```
tc85324, line 107: It would be more idiomatic if these were if, elif, else return NotImplemented. Type hints are missing on the return. Can we also use the Union instead of `|`

tc85324 (changes requested):
A nice addition. There are quite a lot of pyright complaints at the moment though. It would also be nice if something could be done about the auto formatting in some of the places, such as the expected gradients.

tests/unit/test_kernel.py
db091756: It was a way to cast to float without having the line be split over multiple lines but I've changed this to using jnp.full now so not an issue.

tests/unit/test_kernel.py
db091756: Fixed!

coreax/kernel.py
db091756: Yep much nicer, done

coreax/kernel.py
pc532627, line 1: I think it would be worth having an example in the codebase that uses a more complex kernel that is now possible - could you add one - perhaps just an example like `examples/herding_basic.py` but with a kernel that uses both products and additive aspects now, in a new file called `examples/herding_paired_kernel.py` or something similar? We should also then add a corresponding file in `tests/integration/` that runs the test.

coreax/kernel.py
```diff
@@ -365,13 +409,152 @@ def _reshape(x: Array) -> Array:
...
+    :math:`l:\mathbb{R}^d \times \mathbb{R}^d \to \mathbb{R}`, define the additive
+    kernel :math:`p:\mathbb{R}^d \times \mathbb{R}^d \to \mathbb{R}` where
+    :math:`p(x,y) := k(x,y) + l(x,y)`
+
+    :param first_kernel: Instance of class coreax.kernel.Kernel
+    :param second_kernel: Instance of class coreax.kernel.Kernel
+    """
+
+    @override
+    def compute_elementwise(self, x: ArrayLike, y: ArrayLike):
```
pc532627, line 449: Needs output annotation of Array.

coreax/kernel.py
```diff
@@ -365,13 +409,152 @@ def _reshape(x: Array) -> Array:
...
+        return self.first_kernel.compute_elementwise(
+            x, y
+        ) + self.second_kernel.compute_elementwise(x, y)
+
+    @override
+    def grad_x_elementwise(self, x: ArrayLike, y: ArrayLike):
```
pc532627, line 455: Needs output annotation of Array.

coreax/kernel.py
```diff
@@ -365,13 +409,152 @@ def _reshape(x: Array) -> Array:
...
+        return self.first_kernel.grad_x_elementwise(
+            x, y
+        ) + self.second_kernel.grad_x_elementwise(x, y)
+
+    @override
+    def grad_y_elementwise(self, x: ArrayLike, y: ArrayLike):
```
pc532627, line 461: Needs output annotation of Array.

coreax/kernel.py
```diff
@@ -365,13 +409,152 @@ def _reshape(x: Array) -> Array:
...
+        return self.first_kernel.grad_y_elementwise(
+            x, y
+        ) + self.second_kernel.grad_y_elementwise(x, y)
+
+    @override
+    def divergence_x_grad_y_elementwise(self, x: ArrayLike, y: ArrayLike):
```
pc532627, line 467: Needs output annotation of Array.

coreax/kernel.py
```diff
@@ -365,13 +409,152 @@ def _reshape(x: Array) -> Array:
...
+    :math:`l:\mathbb{R}^d \times \mathbb{R}^d \to \mathbb{R}`, define the product kernel
+    :math:`p:\mathbb{R}^d \times \mathbb{R}^d \to \mathbb{R}` where
+    :math:`p(x,y) = k(x,y)l(x,y)`
+
+    :param first_kernel: Instance of class coreax.kernel.Kernel
+    :param second_kernel: Instance of class coreax.kernel.Kernel
+    """
+
+    @override
+    def compute_elementwise(self, x: ArrayLike, y: ArrayLike):
```
pc532627, line 487: Needs output annotation of Array.

coreax/kernel.py
```diff
@@ -365,13 +409,152 @@ def _reshape(x: Array) -> Array:
...
+        if self.first_kernel == self.second_kernel:
+            return self.first_kernel.compute_elementwise(x, y) ** 2
+        return self.first_kernel.compute_elementwise(
+            x, y
+        ) * self.second_kernel.compute_elementwise(x, y)
+
+    @override
+    def grad_x_elementwise(self, x: ArrayLike, y: ArrayLike):
```
pc532627, line 495: Needs output annotation of Array.

coreax/kernel.py
```diff
@@ -365,13 +409,152 @@ def _reshape(x: Array) -> Array:
...
+        return self.first_kernel.grad_x_elementwise(
+            x, y
+        ) * self.second_kernel.compute_elementwise(
+            x, y
+        ) + self.second_kernel.grad_x_elementwise(
+            x, y
+        ) * self.first_kernel.compute_elementwise(x, y)
+
+    @override
+    def grad_y_elementwise(self, x: ArrayLike, y: ArrayLike):
```
pc532627, line 511: Needs output annotation of Array.

coreax/kernel.py
```diff
@@ -365,13 +409,152 @@ def _reshape(x: Array) -> Array:
...
+        return self.first_kernel.grad_y_elementwise(
+            x, y
+        ) * self.second_kernel.compute_elementwise(
+            x, y
+        ) + self.second_kernel.grad_y_elementwise(
+            x, y
+        ) * self.first_kernel.compute_elementwise(x, y)
+
+    @override
+    def divergence_x_grad_y_elementwise(self, x: ArrayLike, y: ArrayLike):
```
pc532627, line 527: Needs output annotation of Array.

tests/unit/test_kernel.py
```diff
@@ -212,22 +215,287 @@ def expected_divergence_x_grad_y(
...
+
+    @pytest.fixture(scope="class")
+    def kernel(self) -> AdditiveKernel:
+        """Return a mocked paired kernel function."""
+        return _MockedPairedKernel(
+            num_points=self.mock_num_points, dimension=self.mock_dimension
+        ).to_additive_kernel()
+
+    @pytest.fixture(params=["floats", "vectors", "arrays"])
+    def problem(self, request, kernel: AdditiveKernel) -> _Problem:
```
pc532627, line 278: `request` is not type annotated here

tests/unit/test_kernel.py
```diff
@@ -212,22 +215,287 @@ def expected_divergence_x_grad_y(
...
+        :math:`p(x,y) := k(x,y) + l(x,y)`
+
+        We consider the simplest possible example of adding two Linear kernels together
+        with the following cases:
+        - `floats`: where x and y are floats
+        - `vectors`: where x and y are vectors of the same size
+        - `arrays`: where x and y are arrays of the same shape
+        """
+        mode = request.param
+        x = 0.5
```
pc532627, line 294: Very minor point, but is there a reason why these are defined outside of the if statement below for floats?

tests/unit/test_kernel.py
```diff
@@ -212,22 +215,287 @@ def expected_divergence_x_grad_y(
...
+
+    @pytest.fixture(scope="class")
+    def kernel(self) -> ProductKernel:
+        """Return a mocked paired kernel function."""
+        return _MockedPairedKernel(
+            num_points=self.mock_num_points, dimension=self.mock_dimension
+        ).to_product_kernel()
+
+    @pytest.fixture(params=["floats", "vectors", "arrays"])
+    def problem(self, request, kernel: ProductKernel) -> _Problem:
```
pc532627, line 388: `request` is not type annotated here

tests/unit/test_kernel.py
```diff
@@ -212,22 +215,287 @@ def expected_divergence_x_grad_y(
...
+        :math:`p(x,y) = k(x,y)l(x,y)`
+
+        We consider the simplest possible example of adding two Linear kernels together
+        with the following cases:
+        - `floats`: where x and y are floats
+        - `vectors`: where x and y are vectors of the same size
+        - `arrays`: where x and y are arrays of the same shape
+        """
+        mode = request.param
+        x = 1.5
```
pc532627, line 404: Same comment as above about defining these outside of the if statement

pc532627 (changes requested):
A few minor style points and a request for an example to be added that showcases this functionality to the user. Generally code looks good.

tests/unit/test_kernel.py
db091756: Just following the structure of the other kernel tests already in, don't think there is any reason

examples/herding_paired_kernel.py
```diff
@@ -0,0 +1,225 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""
+Example coreset generation using randomly generated point clouds.
+
+This example showcases how a coreset can be generated from a dataset containing ``n``
+points sampled from ``k`` clusters in space.
+
+A coreset is generated using kernel herding, with a Squared Exponential kernel. This
```
pc532627, line 21: Update docstring to state we are using a kernel that combines Squared exponential, PCIMQ and Laplace

examples/herding_paired_kernel.py
pc532627, line 1: To reduce the run time of this test in the workflows, I suggest just computing the herding coreset and the MMD for that - this should cut 70% of the run-time off but still show how to use the product and additive kernels

pc532627 (changes requested):
A few very minor points - one about speeding up the workflows which I think will make a big difference

pc532627 (approved):

pc532627 (approved):

PR #621, opened 2024-05-14
Feature/refactor metrics

### LINKED ISSUE #613: Refactor MMD metric to utilise the blocked computation already in `Kernel`
### What's the new feature?

If we generalise the `Kernel.gramian_row_mean` to `Kernel.compute_row_mean`, then all block handling required by MMD will already be handled in `Kernel`.

### What value does this add?

Massively reduces duplication and simplifies the codebase.

_No response_

### Additional context

_No response_

### LINKED ISSUE #631: Compositional/Partitioned Solvers
### What's the new feature?

Create a composite solver that partitions the coreset problem, solves the collection of smaller coreset problems, and then reconstitutes the result as a single coreset.

### What value does this add?

Allows for MapReduce like functionality and provides a means to scale the solving over multiple devices (SPMD acceleration). 

_No response_

### Additional context

_No response_

tc85324:
### PR Type
Closes #613 

- Tests

### Description
Significantly reduces duplication from `coreax.metrics` by using the new `kernel.compute_mean` method. This also allows many of the tests from `test_metrics.py` to be removed as the associated functionality is already tested, for `kernel.compute_mean`, in `test_kernels.py`

Additionaly converts `coreax.util.apply_negative_precision_threshold` to be JIT compatible (allowing `metric.compute` to be JIT compiled).
 
### How Has This Been Tested?
All tests pass

All methods except for `compute` have been removed from `coreax.mdd.MMD`.

### DIFF
--- a/coreax/coreset.py
+++ b/coreax/coreset.py
@@ -31,7 +31,7 @@ class Coreset(eqx.Module, Generic[_Data]):
     with a non-negative (probability) weight :math:`w_i \in \mathbb{R} \ge 0`, there
     exists an implied discrete (probability) measure over :math:`\Omega`
 
-    .. math:
+    .. math::
         \eta_n = \sum_{i=1}^{n} w_i \delta_{x_i}.
 
     If we then specify a set of test-functions :math:`\Phi = {\phi_1, \dots, \phi_M}`,
@@ -39,7 +39,7 @@ class Coreset(eqx.Module, Generic[_Data]):
     "important" properties of the data, then there also exists an implied push-forward
     measure over :math:`\mathbb{R}^M`
 
-    .. math:
+    .. math::
         \mu_n = \sum_{i=1}^{n} w_i \delta_{\Phi(x_i)}.
 
     A coreset is simply a reduced measure containing :math:`\hat{n} < n` updated nodes
@@ -47,12 +47,12 @@ class Coreset(eqx.Module, Generic[_Data]):
     of the coreset :math:`\nu_\hat{n}` has (approximately for some algorithms) the same
     "centre-of-mass" as the push-forward measure for the original data :math:`\mu_n`
 
-    .. math:
+    .. math::
         \text{CoM}(\mu_n) = \text{CoM}(\nu_\hat{n}),
         \text{CoM}(\nu_\hat{n}) = \int_\Omega \Phi(\omega) d\nu_\hat{x}(\omega),
         \text{CoM}(\nu_\hat{n}) = \sum_{i=1}^\hat{n} \hat{w}_i \delta_{\Phi(\hat{x}_i)}.
 
-    .. note:
+    .. note::
         Depending on the algorithm, the test-functions may be explicitly specified by
         the user, or implicitly defined by the algorithm's specific objectives.
 
--- a/coreax/coreset.py
+++ b/coreax/coreset.py
@@ -82,7 +82,7 @@ class Coresubset(Coreset[_Data], Generic[_Data]):
     the reduced measure (the coreset), must be a subset of the support of the original
     measure (the original data), such that
 
-    .. math:
+    .. math::
         \hat{x}_i = x_i, \forall i \in I,
         I \subset \{1, \dots, n\}, text{card}(I) = \hat{n}.
 
--- a/coreax/metrics.py
+++ b/coreax/metrics.py
@@ -60,14 +60,14 @@ class MMD(Metric[_Data]):
     r"""
     Definition and calculation of the (weighted) maximum mean discrepancy metric.
 
-    For a dataset :math:`\mathcal_{D}_1` of ``n`` points in ``d`` dimensions, and
-    another dataset :math:`\mathcal_{D}_2` of ``m`` points in ``d`` dimensions, the
+    For a dataset :math:`\mathcal{D}_1` of ``n`` points in ``d`` dimensions, and
+    another dataset :math:`\mathcal{D}_2` of ``m`` points in ``d`` dimensions, the
     (weighted) maximum mean discrepancy is given by:
 
     .. math:
-        \text{MMD}^2(\mathcal_{D}_1,\mathcal_{D}_2) = \mathbb{E}(k(\mathcal_{D}_1,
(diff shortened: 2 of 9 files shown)


coreax/metrics.py
```diff
@@ -39,472 +39,93 @@
...
 
-    For a dataset of ``n`` points in ``d`` dimensions, :math:`\mathcal_{D}_1`, and
-    another dataset :math:`\mathcal_{D}_2` of ``m`` points in ``d`` dimensions, the
-    maximum mean discrepancy is given by:
+    For a dataset :math:`\mathcal{D}_1` of ``n`` points in ``d`` dimensions, and
+    another dataset :math:`\mathcal{D}_2` of ``m`` points in ``d`` dimensions, the
+    (weighted) maximum mean discrepancy is given by:
 
-    .. math::
+    .. math:
```
tp832944, line 67: Need a double colon at end.

coreax/metrics.py
```diff
@@ -25,12 +25,12 @@
 # Support annotations with | in Python < 3.10
 from __future__ import annotations
```
tp832944, line 26: Given you have largely rewritten this module, would it make sense to ditch this line and add quotes around annotations that are offending in 3.9?

coreax/metrics.py
```diff
@@ -39,472 +39,93 @@
...
-        Calculate maximum mean discrepancy.
+        Compute the (weighted) maximum mean discrepancy.
 
-        If uniform weights are given for datasets ``reference_data`` and
-        ``comparison_data`` standard MMD is calculated. If non-uniform weights are
-        given, weighted MMD is calculated. For both cases, if the size of matrix blocks
-        to process is less than the size of both datasets, and the reference and
-        comparison weights are not uniform in the weighted case, the calculation is done
-        block-wise to limit memory requirements.
+        .. math:
```
tp832944, line 100: Need a double colon at end.

coreax/metrics.py
```diff
@@ -39,472 +39,93 @@
...
-
-        :param reference_points: :math:`n \times d` array of reference data
-        :param comparison_points: An :math:`m \times d` array to compare to
-            ``reference_points``
-        :param block_size: Size of matrix blocks to process
+        :param block_size: Size of matrix blocks to process; a value of :data:`None`
+            sets :math:`B_x = n` and :math:`B_y = m`, effectively disabling the block
+            accumulation; an integer value ``B`` sets :math:`B_y = B_x = B`; a tuple
+            allows different sizes to be specified for ``B_x`` and ``B_y``; to reduce
+            overheads, it is often sensible to select the largest block size which does
```
tp832944, line 114: `that` instead of `which`

coreax/metrics.py
tc85324: I've made this mistake in a few other areas as well (not directly related to the pull), will include those fixes in this pull. Its unfortunate that the docs build without warning, just completely skipping the directives.

tests/unit/test_util.py
```diff
@@ -86,25 +86,22 @@ def test_pairwise_difference(self) -> None:
...
-        A negative precision threshold is given, which should be rejected by the
-        function.
-        """
-        with pytest.raises(ValueError):
-            apply_negative_precision_threshold(x=0.1, precision_threshold=-1e-8)
-
     @pytest.mark.parametrize(
         "value, threshold, expected",
         [
             (-0.01, 0.001, -0.01),
```
tp832944, line 92: Add a negative threshold version of this case.

tp832944 (changes requested):
It's fantastic to see your refactoring in action @tc85324! The code looks a lot more powerful now.

It looks like the second colon missing off Sphinx directives is widespread in your refactored code. Please scan through all files and check and include fixes in this or a new PR.

coreax/metrics.py
tc85324: As I understand, quoting the annotations would be functionally the same as using this import. The alternative would be using the `Union[..., ...]` type, but we don't use this anywhere else in the codebase.

tp832944: In Vanguard, we made the decision that future imports for annotations are a bad idea due to their ability to hide all sorts of evils. Quoting instead provides the exception/ignore on a per-annotation basis rather than per-file. Coreax should move towards this but I'm happy to leave it for now to maintain consistency with the rest of the Coreax codebase.

tests/unit/test_util.py
```diff
@@ -86,32 +80,50 @@ def test_pairwise_difference(self) -> None:
...
+            "tree_length",
+            "above_tree_length",
+        ],
+    )
+    def test_tree_leaves_repeat(self, length: int) -> None:
+        """Test tree_leaves_repeat for various length parameters."""
+        tree = (None, 1)
+        tree_leaves = jtu.tree_leaves(tree, is_leaf=lambda x: x is None)
+        repeated_tree_leaves = tree_leaves_repeat(tree, length)
+        expected_leaves = tree_leaves + [
```
tp832944, line 99: I think it would be better to hardcode `tree_leaves` as `[None, 1]` here and delete line 97. Then, it also tests that `is_leaf` is behaving as expected.

coreax/metrics.py
```diff
@@ -39,472 +39,114 @@
...
-                        * self.kernel.compute(
-                            reference_points[i : i + block_size],
-                            comparison_points[j : j + block_size],
-                        )
-                        * comparison_weights[None, j : j + block_size]
-                    )
-                    weighted_pairwise_distance_sum += pairwise_distances_part.sum()
-
-        return weighted_pairwise_distance_sum
+def _permuted_block_sizes(
```
tp832944, line 135: Add some unit tests for this function. It took me a while to work out the mn spaghetti!

tp832944 (changes requested):

coreax/coreset.py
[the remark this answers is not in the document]
tc85324: Don't know how I missed this!

coreax/metrics.py
tc85324: I've removed this function in favour of `itertools.product` and replaced all the `n` with `x` and all the `m` with `y` as these are much easier to read next to each other.

tp832944 (approved):

PR #626, opened 2024-05-22
Fix/approximate data compatibility

tc85324:
### PR Type

- Bugfix
- Feature

### Description
Since the introduction of `kernel.compute_mean`, the `gramian_row_mean` method is expected to accept `coreax.data.Data` objects. Compatibility is provided by making `coreax.data.Data` a proper JAX  `ArrayLike` type. This allows `Data.data` to be accessed via `jnp.asarray(Data(...))`.
    
In addition, the new `kernel.compute_mean` method allows some duplicated code to be removed from the old approximator implementations.

To reduce duplication, an `as_data` method has been added, and the `kernel.gramian_row_mean` docstring has been simplified (including typehints for block_size).

### How Has This Been Tested?
All tests pass

No

### DIFF
--- a/coreax/approximation.py
+++ b/coreax/approximation.py
@@ -183,7 +183,7 @@ class MonteCarloApproximateKernel(RandomRegressionKernel):
         :return: Approximation of the base kernel's Gramian row-mean
         """
         del kwargs
-        data = jnp.asarray(x)
+        data = jnp.atleast_2d(jnp.asarray(x))
         num_data_points = len(data)
         key = self.random_key
         features_idx = _random_indices(key, num_data_points, self.num_kernel_points - 1)
@@ -223,7 +223,7 @@ class ANNchorApproximateKernel(RandomRegressionKernel):
         :return: Approximation of the base kernel's Gramian row-mean
         """
         del kwargs
-        data = jnp.asarray(x)
+        data = jnp.atleast_2d(jnp.asarray(x))
         num_data_points = len(data)
         features = jnp.zeros((num_data_points, self.num_kernel_points))
         features = features.at[:, 0].set(self.base_kernel.compute(data, data[0])[:, 0])
@@ -279,7 +279,7 @@ class NystromApproximateKernel(RandomRegressionKernel):
         :return: Approximation of the base kernel's Gramian row-mean
         """
         del kwargs
-        data = jnp.asarray(x)
+        data = jnp.atleast_2d(jnp.asarray(x))
         num_data_points = len(data)
         feature_idx = _random_indices(
             self.random_key, num_data_points, self.num_kernel_points
--- a/coreax/approximation.py
+++ b/coreax/approximation.py
@@ -27,10 +27,9 @@ all functionality provided through composition with a ``base_kernel``, they can 
 freely used in any place where a standard :class:`~coreax.kernel.Kernel` is expected.
 """
 
-from __future__ import annotations
-
 from collections.abc import Callable
 from functools import partial
+from typing import Union
 
 import jax
 import jax.numpy as jnp
@@ -173,7 +172,7 @@ class MonteCarloApproximateKernel(RandomRegressionKernel):
     :param num_train_points: Number of training points used to fit kernel regression
     """
 
-    def gramian_row_mean(self, x: ArrayLike | Data, **kwargs) -> Array:
+    def gramian_row_mean(self, x: Union[ArrayLike, Data], **kwargs) -> Array:
         r"""
         Approximate the Gramian row-mean by Monte-Carlo sampling.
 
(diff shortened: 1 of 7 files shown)


coreax/approximation.py
```diff
@@ -180,7 +183,7 @@ def gramian_row_mean(self, x: ArrayLike, **kwargs) -> Array:
         :return: Approximation of the base kernel's Gramian row-mean
         """
         del kwargs
```
pc532627, line 185: Just as a sense check, do you know when this was introduced, and what it needs doing at the top of all of these methods? (deleting the `kwards`)?

coreax/approximation.py
```diff
@@ -180,7 +183,7 @@ def gramian_row_mean(self, x: ArrayLike, **kwargs) -> Array:
...
-        data = jnp.atleast_2d(x)
+        data = jnp.asarray(x)
```
pc532627, line 186: In all previous cases, we were assuming data was handled as a 2D JAX array (with `jnp.atleast_2d`). I assume `jnp.asarray` is needed for compatibility with `x` if `x` is of type `Data`? Can we not keep the assumption that everything is worked with as a 2D array?

coreax/data.py
```diff
@@ -207,19 +207,23 @@ class Data(eqx.Module):
...
     def __init__(
         self,
-        data: Shaped[ArrayLike, " n d"],
+        data: Shaped[ArrayLike, " n *d"],
         weights: Shaped[ArrayLike, " n"] | None = None,
     ):
         """Initialise Data class."""
-        self.data = jnp.atleast_2d(data)
-        n = self.data.shape[0]
+        self.data = jnp.asarray(data)
```
pc532627, line 219: Same comment about `jnp.asarray` vs `jnp.atleast_2d` - is this needed and has the code been tested if 1d arrays are passed?

pc532627 (changes requested):

coreax/approximation.py
tc85324: The kwargs parameter is for Liskov, the del is to supress pylint's complaint of an unused variable.

coreax/approximation.py
tc85324: We only test for 2D at the moment. I don't like the usage of `atleast_2d` as it is not explicit to the user where this dimension will be added (the leading axis). If the user passes an n-vector (believing this represents n scalar observations) then `atleast_2d` will silently give them the wrong result. This is because `atleast_2d` converts the vector to an array with shape `(1, n)`, rather than `(n, 1)`. 

The difficulty is that different users may assume different behaviour for vectors. Some may (correctly in this case) assume an n-vector represents a single n-dimensional observation, whereas others may (incorrectly in this case) assume an n-vector represents n 1-dimensional observations.

Because of this, I think we only have three options: 
1. throw an error when passed an array that is not 2D `if jnp.ndim(x) != 2: raise ValueError("Must be 2D")`
2. allow k-dimensional arrays and let JAX handle any resulting incompatible mathematical operations.
3. use beartype and jaxtyping to provide (optional) runtime type checking of the array dimensions.

If we start using jaxtyping everywhere then the expectations on user data become clearer before runtime (at programming time).

Addressing this issue properly across the entire codebase is out of scope for this pull, so I would prefer to keep this as is, or just wrap the `asarrays` in an `atleast_2d` just to preserve the previous behaviour (whichever you would prefer).

pc532627: @tc85324 Thanks for highlighting the options. I agree that it is not explicit to the user what will happen with particular inputs and could lead to confusion. My proposal is to wrap the `asarrays` in an `atleast_2d` call to preserve the previous behavior for now, but I will also create a ticket that details this discussion and in resolving that, we do one of the three listed options and remove the wrapped `atleast_2d` calls. How does that sound?

pc532627: For reference this is now discussed in issue #630

coreax/data.py
pc532627: @tc85324 this needs a `atleast_2d` wrapper I believe

pc532627 (changes requested):
I think only 1 line has a change required now

coreax/data.py
tc85324: The type hint indicates that d is optional, so will be treated as an n-vector. In this case we have logical consistency as weights is expected to be an n-vector and the docstrings make clear that each weight is associated with a leading index of the data. If you have atleast_2d here, you end up with the confusing scenario where you pass an `n-vector` of data but are only expected to pass a single scalar weight. In addition, there is no reason why data should have to be 2d here, whereas at the point of use (in the gramian_row_mean of the approximators) the mathematical operations doe require 2d.

pc532627 (approved):

Issue #590, opened 2024-04-29
Approximate inversion methods


db091756:
### What's the new feature?

Various algorithms in [1] exist to produce approximate eigen-decompositions of matrices. 

That is, given some hyper-parameter $r > 0$, one can derive an approximate eigen-decomposition of a matrix $A \in \mathbb{R}^{n \times n}$ such that we have $A \approx U\Lambda U^T$ where $\Lambda \in \mathbb{R}^{r \times r}$ is the diagonal matrix of eigenvalues and $U \in \mathbb{R}^{n \times r}$ is the ortho-normal matrix of eigenvectors. 

From this, it follows that we have  $A^{-1} \approx U\Lambda^{-1} U^T$.

The algorithms that I think we should consider adding are:

- ~Algorithm 5.3. (direct algorithm for Hermitan matrices, most simple to implement) #638~
- ~Algorithm 5.5 (Nystrom based method, good for positive semi-definite matrices such as our kernel matrices) #640~
- Algorithm 5.6. (one pass algorithm for memory management) #641

I suggest that implementing 5.3. and 5.5. should be the priority, with 5.6. next.

References
[1] - https://arxiv.org/abs/0909.4061

### What value does this add?

In [1] they show that their approximate algorithms are $\mathcal{O}(rn^2)$ with $r \ll n$ which is much faster than the usual $\mathcal{O}(n^3)$ cost of matri inversion via eigen-decomposition. This will therefore make the CMMD related algorithms much faster.

tp832944:
@db091756 what is blocking this? Please tag the relevant tickets in the description or as a comment.

db091756:
Currently blocked by refactoring of reduction.py, @tc85324 is working on this currently I believe. Unsure if there is a ticket floating around for this?

be904619:
@db091756 update this ticket to indicate its scope is about choosing which methods we will add, and then add a separate ticket for each of these.

db091756:
Experimentation with an implementation of algorithm 5.5. in the linked branch of #640 seems to suggest that this method is not accurate enough for our purposes as kernel matrices do not satisfy numerical positive semi defininiteness to the required degree. 

Algorithm 5.6. will not reduce memory cost by an order of magnitude as we have to compute the kernel matrix anyway at a memory cost of n^2

PR #637, opened 2024-05-30
Feature/approximate inversion (Alg 5.3)

### LINKED ISSUE #638: Implement Algorithm 5.3.
### What's the new feature?

Implement Algorithm 5.3. described in #590.

### What value does this add?

Faster matrix inversion.

_No response_

### Additional context

_No response_

db091756:
Closes #638 

### PR Type

- Feature
- Tests

### Description

Implements algorithm 5.3. from #590.

### How Has This Been Tested?

Additional unit tests pass

(Write your answer here.)

### Screenshots

(Write your answer here.)


coreax/util.py
```diff
@@ -244,6 +244,45 @@ def solve_qp(kernel_mm: ArrayLike, gramian_row_mean: ArrayLike, **osqp_kwargs) -
...
+        to compute this in JAX requires the 'identity' array to be a matrix of zeros
+        except for ones on the diagonal up to the size of the non-zero block.
+
+    :param array: Array to be inverted
+    :param regularisation_parameter: Regularisation parameter for stable inversion of
+        array, negative values will be converted to positive
+    :param identity: Block "identity" matrix
+    :param rcond: Cut-off ratio for small singular values of a. For the purposes of rank
+        determination, singular values are treated as zero if they are smaller than
+        rcond times the largest singular value of a. The default value of None will use
```
pc532627, line 270: Same comment on definition of `a` here as in previous comment

coreax/util.py
```diff
@@ -244,6 +244,45 @@ def solve_qp(kernel_mm: ArrayLike, gramian_row_mean: ArrayLike, **osqp_kwargs) -
...
+        the machine precision multiplied by the largest dimension of the array.
+        An alternate value of -1 will use machine precision.
+    :return: Inverse of regularised array
+    """
+    if rcond is not None:
+        if rcond < 0 and rcond != -1:
+            raise ValueError("rcond must be non-negative, except for value of -1")
```
pc532627, line 277: `rcond` should be contained within single quotes

coreax/util.py
```diff
@@ -244,6 +244,45 @@ def solve_qp(kernel_mm: ArrayLike, gramian_row_mean: ArrayLike, **osqp_kwargs) -
...
+    if array.shape != identity.shape:
+        raise ValueError("Leading dimensions of array and identity must match")
```
pc532627, line 279: `array` should be contained within single quotes

pc532627: As-well as `identity`

tests/unit/test_approximation.py
```diff
@@ -270,3 +278,171 @@ def test_invalid_kernel(
...
+    @abstractmethod
+    def approximator(self) -> _RegularisedInverseApproximator:
+        """Abstract pytest fixture which initialises an inverse approximator."""
+
+    @abstractmethod
+    def problem(self) -> _InversionProblem:
+        """Abstract pytest fixture which returns a problem for inverse approximation."""
+
+    def test_approximation_accuracy(
+        self, problem, approximator: _RegularisedInverseApproximator
```
pc532627, line 308: `problem` does not have a type annotation here

tests/unit/test_approximation.py
```diff
@@ -270,3 +278,171 @@ def test_invalid_kernel(
...
+            random_key=jr.key(random_seed),
+            oversampling_parameter=100,
+            power_iterations=1,
+            rcond=None,
+        )
+
+    @override
+    @pytest.fixture(scope="class")
+    def problem(self):
+        r"""Define data shared across tests."""
```
pc532627, line 349: Don't need the `r` at the start of this docstring

tests/unit/test_approximation.py
pc532627, line 1: Checking coverage, I think you need a unit test that covers calling the `init` for `RandomisedEigendecompositionApproximator` when the condition `if self.rcond < 0 and self.rcond != -1:` is both True & False, not just one case as current

pc532627: Also, in `RandomisedEigendecompositionApproximator.approximate`, you have the block of code:

```
# Set rcond parameter if not given 

n = kernel_gramian.shape[0] 
machine_precision = jnp.finfo(kernel_gramian.dtype).eps 
if self.rcond is None:
    rcond = machine_precision * n 
elif self.rcond == -1: 
    rcond = machine_precision 
else: 
    rcond = self.rcond
```

I don't think the `elif` and `else` cases are hit with a test

pc532627 (changes requested):
Only very minor comments, code looks good @db091756. As a side note, when the CMMD PR is merged in, we should ensure there is an end-to-end test showcasing this approximator in use.

coreax/approximation.py
```diff
@@ -292,3 +293,209 @@ def gramian_row_mean(self, x: Union[ArrayLike, Data], **kwargs) -> Array:
...
+            of array, negative values will be converted to positive
+        :param identity: Block identity matrix
+        :return: Approximation of the kernel matrix inverse
+        """
+
+
+def randomised_eigendecomposition(
+    random_key: KeyArrayLike,
+    array: Array,
+    oversampling_parameter: int = 10,
```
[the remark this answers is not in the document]
db091756: From the Halko paper, where `p` is the `oversampling_parameter`

"""
_Just how many extra samples do we need? Remarkably, for certain types of
random sampling schemes, the failure probability decreases superexponentially with
the oversampling parameter p. As a practical matter, setting p = 5 or p = 10
often gives superb results. This observation is one of the principal facts supporting
the randomized approach to numerical linear algebra._
"""

Also see:

"""
_The number of columns ℓ that the algorithm needs to reach this threshold is usually slightly larger than the minimal rank k of the smallest basis that verifies (4.2; **here 4.2. is the error the authors consider for the matrix that has a similar range to the target array**). We refer to this discrepancy p = ℓ − k as the oversampling parameter. The size of the oversampling parameter depends on several factors:

1. The matrix dimensions. Very large matrices may require more oversampling.
2. The singular spectrum. The more rapid the decay of the singular values, the less oversampling is needed. In the extreme case that the matrix has exact rank k, it is not necessary to oversample._ 
"""

Some experimentation led me to use larger values of `p` for the tests as the kernel matrix I used does not always have a rapidly decaying singular spectrum. The `oversampling_parameter ` should also not be any larger than the size of the matrix as then you are gaining nothing by this approach.

It could be worth implementing Algorithm 4.2 from the paper as this avoids leaving the choice of `oversampling_parameter` to the user.

On the `power_iterations` front Halko has this to say:
"""
Unfortunately, the simplest version of this scheme is inadequate in many
applications because the singular spectrum of the input matrix may decay slowly. To
address this difficulty, we incorporate q steps of a power iteration, where q = 1 or
q = 2 usually suffices in practice.
"""

coreax/approximation.py
```diff
@@ -292,3 +293,209 @@ def gramian_row_mean(self, x: Union[ArrayLike, Data], **kwargs) -> Array:
...
+
+    # Power iterations for improved accuracy
+    for _ in range(power_iterations):
+        approximate_range_ = array.T @ q
+        q_, _ = jnp.linalg.qr(approximate_range_)
+        approximate_range = array @ q_
+        q, _ = jnp.linalg.qr(approximate_range)
+
+    # Form the low rank array, compute its exact eigendecomposition and ortho-normalise
+    # the eigenvectors.
```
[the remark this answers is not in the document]
db091756: I thought we put full-stops for comments over multiple lines?

coreax/approximation.py
```diff
@@ -292,3 +293,209 @@ def gramian_row_mean(self, x: Union[ArrayLike, Data], **kwargs) -> Array:
...
+    does such an approximation using a randomised eigendecomposition. Further details
+    can be found in (:cite:`halko2009randomness` Algorithm 4.4. and 5.3).
+
+    :param random_key: Key for random number generation
+    :param oversampling_parameter: Number of random columns to sample; the larger the
+        oversampling_parameter, the more accurate, but slower the method will be
+    :param power_iterations: Number of power iterations to do; the larger the
+        power_iterations, the more accurate, but slower the method will be
+    :param rcond: Cut-off ratio for small singular values of a. For the purposes of
+        rank determination, singular values are treated as zero if they are smaller than
```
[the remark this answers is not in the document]
db091756: Should have been `array`, good spot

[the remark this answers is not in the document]
db091756: Or the gramian rather

coreax/approximation.py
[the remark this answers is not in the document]
db091756: There is an element of diminishing returns, especially with `power_iterations`, it is one of the more costly parts of the algorithm and once you have reduced the error to machine precision it is pointless to keep beating the horse

tests/unit/test_approximation.py
db091756: Hopefully added some tests that cover these cases

coreax/approximation.py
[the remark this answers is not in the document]
pc532627: Checked the style guide and you are correct - sorry about that, no change needed.

tests/unit/test_approximation.py
pc532627: All now hit, thanks

coreax/approximation.py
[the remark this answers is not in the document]
pc532627: Thanks for the detail here - the default values seem sensible in this case and since the reference to the paper is there in the docstring, I'm happy to leave as is and the reader can refer to the paper if they want to change these.

pc532627 (approved):

PR #643, opened 2024-05-30
feat: added function to sample an array of batched indices

db091756:
### PR Type

- Feature
- Tests

### Description

Adds a function to sample an array of batch indices for coresubset methods.

### How Has This Been Tested?

 Additional unit tests pass.

(Write your answer here.)

### Screenshots

(Write your answer here.)

### DIFF
--- a/coreax/util.py
+++ b/coreax/util.py
@@ -33,6 +33,7 @@ import equinox as eqx
 import jax.numpy as jnp
 import jax.tree_util as jtu
 from jax import Array, block_until_ready, jit, vmap
+from jax.random import permutation
 from jax.typing import ArrayLike
 from jaxopt import OSQP
 from typing_extensions import TypeAlias, deprecated
@@ -242,6 +243,41 @@ def solve_qp(kernel_mm: ArrayLike, gramian_row_mean: ArrayLike) -> Array:
         params_obj=(q_array, c), params_eq=(a_array, b), params_ineq=(g_array, h)
     ).params
     return sol.primal
+
+
+def sample_batch_indices(
+    random_key: KeyArrayLike,
+    data_size: int,
+    batch_size: int,
+    num_batches: int,
+) -> ArrayLike:
+    """
+    Sample an array of indices of size batch_size x num_batches.
+
+    Each column of the sampled array will contain unique elements. The largest possible
+    index is dictated by data_size.
+
+    :param random_key: Key for random number generation
+    :param data_size: Size of the data we wish to sample from
+    :param batch_size: Size of the batch we wish to sample
+    :param num_batches: Number of batches to sample
+    :return: Array of batch indices of size `batch_size` x `num_batches`
+    """
+    if data_size < batch_size:
+        raise ValueError("data_size must be greater than or equal to batch_size")
+    if (data_size <= 0.0) or not isinstance(data_size, int):
+        raise ValueError("data_size must be a positive integer")
+    if (batch_size <= 0.0) or not isinstance(batch_size, int):
+        raise ValueError("batch_size must be a positive integer")
+    if (num_batches <= 0.0) or not isinstance(num_batches, int):
+        raise ValueError("num_batches must be a positive integer")
+
+    return permutation(
+        key=random_key,
+        x=jnp.tile(jnp.arange(data_size, dtype=jnp.int32), (num_batches, 1)).T,
+        axis=0,
+        independent=True,
+    )[:batch_size, :]
 
 
 def jit_test(
--- a/tests/unit/test_util.py
+++ b/tests/unit/test_util.py
@@ -26,6 +26,7 @@ import equinox as eqx
 import jax.numpy as jnp
 import numpy as np
 import pytest
+from jax.random import key
 from scipy.stats import ortho_group
 
 from coreax.util import (
@@ -34,6 +35,7 @@ from coreax.util import (
     difference,
     jit_test,
     pairwise,
+    sample_batch_indices,
     solve_qp,
     squared_distance,
     tree_leaves_repeat,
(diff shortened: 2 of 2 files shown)


tc85324 (commented):
Maybe I'm missing something, but can't we do this with:
```python
jr.choice(random_key, data_size, (num_batches, batch_size))
```

db091756:
> Maybe I'm missing something, but can't we do this with:
> 
> ```python
> jr.choice(random_key, data_size, (num_batches, batch_size))
> ```

It's not quite the same as the current code ensures that each column (corresponding to each batch of indices for each iteration of the algorithm) contains unique indices to avoid wasted work. 

image

The above image shows what I mean. The second column of the suggested code would result in wasted work, whereas sample_batch_indices gives columns with unique elements.

coreax/util.py
```diff
@@ -244,6 +245,41 @@ def solve_qp(kernel_mm: ArrayLike, gramian_row_mean: ArrayLike, **osqp_kwargs) -
...
+    if data_size < batch_size:
+        raise ValueError("data_size must be greater than or equal to batch_size")
+    if (data_size <= 0.0) or not isinstance(data_size, int):
+        raise ValueError("data_size must be a positive integer")
+    if (batch_size <= 0.0) or not isinstance(batch_size, int):
+        raise ValueError("batch_size must be a positive integer")
+    if (num_batches <= 0.0) or not isinstance(num_batches, int):
+        raise ValueError("num_batches must be a positive integer")
+
+    return permutation(
```
tc85324, line 275: I think this would be more idiomatic, and the logic easier to follow, if this were written:
```python
batch_keys = jr.split(key, num_batches)
batch_permutation = jax.vmap(jr.permutation, in_axes=(0, None))
return batch_permutation(batch_keys, data_size)
```
We don't need to set `independant=True`, because we only consider one vector at a time (inside the vmap).

coreax/util.py
```diff
@@ -244,6 +245,41 @@ def solve_qp(kernel_mm: ArrayLike, gramian_row_mean: ArrayLike, **osqp_kwargs) -
...
+    Each column (batch) of the sampled array will contain unique elements. The largest
+    possible index sampled is dictated by data_size.
+
+    :param random_key: Key for random number generation
+    :param data_size: Size of the data we wish to sample from
+    :param batch_size: Size of the batch we wish to sample
+    :param num_batches: Number of batches to sample
+    :return: Array of batch indices of size `batch_size` x `num_batches`
+    """
+    if data_size < batch_size:
```
tc85324, line 266: I believe the approach we have been taking so far is to only raise custom messages when the underlying code doesn't raise anything helpful/obvious to the user, and to be as permissive as possible.

By this I mean that, for example: if `data_size=1.0`, it will not pass the isinstance check, but it would pass conversion to an int, `int(1.0)`. While the type hint restricts to integers, we can reasonably accept any value that can be cast to an integer. The caveat here is that an 'invalid' value of say '1.5' will be silently cast to the integer `1.0`, which may be unexpected. However, the type hint is already indicating that behaviour is undefined for anything that isn't an `int`, we're just trying to be as lenient as possible. This is the same reason why we use try-except in many places, over explicit type checks.

tests/unit/test_util.py
```diff
@@ -187,6 +189,50 @@ def test_solve_qp_invalid_gramian_row_mean(self) -> None:
...
+    def test_sample_batch_indices(
+        self,
+        data_size: int,
+        batch_size: int,
+        num_batches: int,
+    ) -> None:
+        """
+        Test sample_batch_indices for valid input parameters.
+        """
+        with pytest.raises(ValueError):
```
tc85324, line 228: This test should also check the desired/required properties of the result (uniqueness of the indices, etc...)

tc85324 (changes requested):
Thanks for your explanation, I understand the utility now! My comments are more stylistic than anything; the test probably should be stricter though.

tests/unit/test_util.py
db091756: Added uniqueness test

coreax/util.py
db091756: Makes sense, removed the unnecessary checks. Left the ones in that don't raise errors but behave in a weird way. Removed the corresponding tests

coreax/util.py
db091756: Much nicer! Updated the code above to ensure the array comes out the right size.

tests/unit/test_util.py
```diff
@@ -235,6 +259,22 @@ def test_sample_batch_indices(
...
+        data_size = 100
+        batch_size = 50
+        num_batches = 10
+        batch_indices = sample_batch_indices(
+            random_key=key(0),
+            data_size=data_size,
+            batch_size=batch_size,
+            num_batches=num_batches,
+        )
+        for row in batch_indices:
```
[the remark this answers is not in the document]
tc85324: We currently have an intra-row check (uniqueness of indices within the row), but not an inter-row uniqueness check (uniqueness between the rows). I assume this is because inter-row uniqueness is not guaranteed. I.E we would accept `[1,2],[1,2],[2,1],[1,2]` for four batches of size two, where data_size is also two. My only concern is that the test currently permits a function that generates the same order of indices for every batch.

I think the probability of sampling any specific (ordered) batch of indices is `1 / math.perm(data_size, batch_size)`; with the parameters in this test, the probability is on the order of 1e-94 (the parameters could be chosen to make this probability arbitrarily small). I believe this allows us to reasonably expect all rows to be unique (perhaps leave a comment in the test explaining why this is expected).

The test for the above could be something like `assert jnp.unique(batch_indices, axis=0).shape[0] == num_batches` . This will implicitly check the shape of `batch_indices` if you don't want a dedicate shape assertion.

tc85324 (changes requested):
Apologies for taking time to get back to this. I'll be finishing existing reviews and PRs at a slower pace, but still intend to get them all completed. Thank you for responding to my previous comments. Most of my new comments are somewhat pedantic nitpicks (sorry!); the biggest issue is what I believe to be the accidental inclusion of regularised array inversion in this PR.

coreax/util.py
```diff
@@ -245,39 +245,72 @@ def solve_qp(kernel_mm: ArrayLike, gramian_row_mean: ArrayLike, **osqp_kwargs) -
     return sol.primal
 
 
+def invert_regularised_array(
```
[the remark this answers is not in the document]
db091756: This has been merged in now

tests/unit/test_util.py
```diff
@@ -191,34 +192,57 @@ def test_solve_qp_invalid_gramian_row_mean(self) -> None:
...
+        """Test that `invert_regularised_array` handles invalid inputs."""
+        with pytest.raises(ValueError):
+            invert_regularised_array(
+                array=array,
+                regularisation_parameter=1e-6,
+                identity=identity,
+                rcond=rcond,
+            )
+
+    def test_invert_regularised_array(self) -> None:
```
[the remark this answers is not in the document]
db091756: Merged in now

tests/unit/test_util.py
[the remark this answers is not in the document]
db091756: Added tests that cover these cases

tc85324 (changes requested):
This looks good! One minor nitpick on the function docstring then I'm happy to merge

coreax/util.py
```diff
@@ -245,39 +245,72 @@ def solve_qp(kernel_mm: ArrayLike, gramian_row_mean: ArrayLike, **osqp_kwargs) -
...
+    )[0]
+
+
 def sample_batch_indices(
     random_key: KeyArrayLike,
     data_size: int,
     batch_size: int,
     num_batches: int,
-) -> ArrayLike:
+) -> Array:
```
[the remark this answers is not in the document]
db091756: Good spot! Fixed

tc85324 (approved):

PR #646, opened 2024-06-04
refactor: removed pipe operator

db091756:
### PR Type

### Description

Replace | operator with Union and used Optional type.

(Write your answer here.)

### Screenshots

(Write your answer here.)

### DIFF
--- a/coreax/kernel.py
+++ b/coreax/kernel.py
@@ -51,7 +51,7 @@ automatic differentiated default.
 from abc import abstractmethod
 from collections.abc import Callable
 from math import ceil
-from typing import TypeVar, Union
+from typing import Optional, TypeVar, Union
 
 import equinox as eqx
 import jax
@@ -235,7 +235,7 @@ class Kernel(eqx.Module):
         self,
         x: Union[ArrayLike, Data],
         *,
-        block_size: Union[int, None, tuple[Union[int, None, int, None]]] = None,
+        block_size: Union[int, None, tuple[Optional[int], Optional[int]]] = None,
         unroll: Union[int, bool, tuple[Union[int, bool], Union[int, bool]]] = 1,
     ):
         r"""
@@ -255,9 +255,9 @@ class Kernel(eqx.Module):
         self,
         x: Union[ArrayLike, Data],
         y: Union[ArrayLike, Data],
-        axis: Union[int, None] = None,
+        axis: Optional[int] = None,
         *,
-        block_size: Union[int, None, tuple[Union[int, None, int, None]]] = None,
+        block_size: Union[int, None, tuple[Optional[int], Optional[int]]] = None,
         unroll: Union[int, bool, tuple[Union[int, bool], Union[int, bool]]] = 1,
     ) -> Array:
         r"""
@@ -342,7 +342,7 @@ class Kernel(eqx.Module):
 
 
 def _block_data_convert(
-    x: Union[ArrayLike, Data], block_size: Union[int, None]
+    x: Union[ArrayLike, Data], block_size: Optional[int]
 ) -> tuple[Array, int]:
     """Convert 'x' into padded and weight normalized blocks of size 'block_size'."""
     x = as_data(x).normalize(preserve_zeros=True)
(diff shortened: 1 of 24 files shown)


tc85324 (commented):
Good use of the `Optional` type, I had forgotten all about this! Hence, lots of `Union[..., None]`.

I do think the `Optional` use in the `block_size` and `unroll` type hints needs changing though. For example, with `block_size: Union[int, None, tuple[Optional[int], Optional[int]]]`, if we pass a tuple, it isn't optional to pass an `int`. The tuple should be of type `int` or `None`. It is, however, optional to pass an integer or a tuple, so I think the correct hint is `block_size: Optional[int, tuple[Union[int, None], Union[int, None]]] = None`.

coreax/kernel.py
```diff
@@ -391,7 +390,7 @@ def slice_sum(accumulated_sum: Array, y_block: Data) -> tuple[Array, Array]:
 
 
 def _block_data_convert(
-    x: Union[ArrayLike, Data], block_size: Union[int, None]
+    x: Union[ArrayLike, Data], block_size: Optional[int]
```
tc85324, line 393: While it isn't obvious or clear from https://docs.python.org/3/library/typing.html#typing.Optional, I believe the only time `Optional` should be used is if we have a default assignment of `None`. In all other cases, like here, I think `Union[int, None]` is more appropriate. Otherwise, the current type hint reads as if `block_size` is optional, even though it is not. I think this is why the `int | None` syntax, supported in newer versions of python, is better (there is no ambiguity).

coreax/metrics.py
```diff
@@ -91,8 +88,8 @@ def compute(
         reference_data: _Data,
         comparison_data: _Data,
         *,
-        block_size: int | None | tuple[int | None, int | None] = None,
-        unroll: tuple[int | bool, int | bool] = (1, 1),
+        block_size: Union[int, None, tuple[Union[int, None], Union[int, None]]] = None,
```
tc85324, line 91: I guess the choice not to use `Optional` around a `Union` of the `int` and `tuple` is to avoid excessive nesting?

coreax/metrics.py
```diff
@@ -91,8 +88,8 @@ def compute(
...
+        unroll: tuple[Union[int, bool], Union[int, bool]] = (1, 1),
```
tc85324, line 92: The type hint and the default here should have been updated in some previous commit to match that of `Kernel.compute_mean`, which is `Union[int, bool, tuple[Union[int, bool], Union[int, bool]]] = 1`. Please can you update to match here

documentation/source/conf.py
```diff
@@ -193,7 +185,7 @@
 }
 
 
-def typehints_formatter(annotation: Any, config: sphinx.config.Config) -> str | None:
+def typehints_formatter(annotation: Any, config: sphinx.config.Config) -> Optional[str]:
```
tc85324, line 188: Not sure how I feel about this being optional. Would probably prefer `Union` just so all uses of optional are those with a default value of `None`

tc85324 (changes requested):
I've not checked for any missed hints, but most of the changes in the PR look good! I've left a few comments about my opinion on using `Optional` but I'm happy to have my mind changed.

coreax/metrics.py
db091756: Yep it seemed excessive!

coreax/metrics.py
db091756: Fixed

coreax/kernel.py
```diff
@@ -290,7 +290,6 @@ def gramian_row_mean(
         r"""
         Compute the (blocked) row-mean of the kernel's Gramian matrix.
 
-        A convenience method for calling meth:`compute_mean`. Equivalent to the call
```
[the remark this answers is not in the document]
db091756: Nope, fixed!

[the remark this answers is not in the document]
tc85324: There is a typo in the original line. The `meth` directive is missing the leading colon, should be:
"calling :meth:`compute_mean`".

tc85324 (changes requested):
This PR is a nice improvement to the type hinting consistency! Happy to approve once the docstring typo in the remaining unresolved comment is handled.

coreax/kernel.py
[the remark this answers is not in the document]
db091756: Fixed

tc85324 (approved):

PR #661, opened 2024-06-14
refactor: augmented unroll parameter to be consistent with block size

### LINKED ISSUE #648: Treat Unroll parameters like Block Size parameters in MMD
### What's the new feature?

The unroll parameters passed to each `compute_mean` call in `MMD.compute` should be augmented. If the user passes `unroll` as a tuple `(unroll_x, unroll_y)`, then we should augment this tuple for each call to `compute_mean`
- `compute_mean(x, x, unroll=(unroll_x, unroll_x))`
- `compute_mean(y, y, unroll=(unroll_y, unroll_y))`
- `compute_mean(x, y, unroll=(unroll_x, unroll_y))`

### What value does this add?

This provides consistent handling with that of `block_size`.

_No response_

### Additional context

_No response_

db091756:
### PR Type

### Description

Closes #648
### How Has This Been Tested?

All existing unit tests pass.

(Write your answer here.)

### Screenshots

(Write your answer here.)

### DIFF
--- a/coreax/metrics.py
+++ b/coreax/metrics.py
@@ -119,14 +119,15 @@ class MMD(Metric[_Data]):
         del kwargs
         _block_size = coreax.util.tree_leaves_repeat(block_size, 2)
         bs_xx, bs_xy, _, bs_yy = tuple(product(_block_size, repeat=len(_block_size)))
-        u_x, u_y = coreax.util.tree_leaves_repeat(unroll, 2)
+        _unroll = coreax.util.tree_leaves_repeat(unroll, 2)
+        u_xx, u_xy, _, u_yy = tuple(product(_unroll, repeat=len(_unroll)))
         # Variable rename allows for nicer automatic formatting
         x, y = reference_data, comparison_data
-        k_xx_mean = self.kernel.compute_mean(x, x, block_size=bs_xx, unroll=(u_x, u_x))
-        k_yy_mean = self.kernel.compute_mean(y, y, block_size=bs_yy, unroll=(u_y, u_y))
-        k_xy_mean = self.kernel.compute_mean(x, y, block_size=bs_xy, unroll=(u_x, u_y))
+        kernel_xx_mean = self.kernel.compute_mean(x, x, block_size=bs_xx, unroll=u_xx)
+        kernel_yy_mean = self.kernel.compute_mean(y, y, block_size=bs_yy, unroll=u_yy)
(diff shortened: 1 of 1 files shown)


coreax/metrics.py
```diff
@@ -119,13 +119,14 @@ def compute(
         del kwargs
         _block_size = coreax.util.tree_leaves_repeat(block_size, 2)
         bs_xx, bs_xy, _, bs_yy = tuple(product(_block_size, repeat=len(_block_size)))
+        u_x, u_y = coreax.util.tree_leaves_repeat(unroll, 2)
```
tc85324, line 122: For purely stylistic reasons, I think it would be better if we exactly matched the block_size case. 
```python
_unroll = coreax.util.tree_leaves_repeat(unroll, 2)
u_xx, u_xy, _, u_yy = tuple(product(_block_size, repeat=len(_unroll)))
```
Not that it really matters, but from the perspective of compilation time, we could slightly reduce compilation time here by using a scan, which also includes the `compute_mean` lines. Something like:
```python
_block_size = coreax.util.tree_leaves_repeat(block_size, 2)
_unroll = coreax.util.tree_leaves_repeat(unroll, 2)
_data = reference_data, comparison_data
indices = [(0,0),(0,1),(1,1)] # i_xx, i_xy, i_yy

def _mean(_, val):
    i_x, i_y = val
    (x, y), bs, un = [(z[i_x], z[i_y]) for z in (_data, _block_size, _unroll)]
    return _, self.kernel.compute_mean(x, y, block_size=bs, unroll=un)

_, [k_xx_mean, k_xy_mean, k_yy_mean] = jax.lax.scan(_mean, None, indices)
squared_mmd_threshold_applied = coreax.util.apply_negative_precision_threshold(
    k_xx_mean + k_yy_mean - 2 * k_xy_mean,
    self.precision_threshold,
)
```
If you want to time this, I would be interested to see if there is an appreciable difference. If not, then I'm happy to leave this as a future ticket. Note: assume compilation time is roughly the difference in runtime between the first and second call of a JIT'ed function.

tc85324 (changes requested):

coreax/metrics.py
db091756: I've made that stylistic change, probably worth making a new ticket for the compilation time Q

tc85324 (approved):

PR #698, opened 2024-07-10
fix negative weights

db091756:
### PR Type

- Bugfix

### Description

Ensures weights are greater than or equal to zero and sum to one. Fixes bug detailed in listed issue.

### How Has This Been Tested?

All tests pass.

(Write your answer here.)

### Screenshots

(Write your answer here.)

### DIFF
--- a/coreax/util.py
+++ b/coreax/util.py
@@ -202,7 +202,6 @@ def pairwise_difference(x: ArrayLike, y: ArrayLike) -> Array:
 def solve_qp(
     kernel_mm: ArrayLike,
     gramian_row_mean: ArrayLike,
-    precision_threshold: float = 1e-12,
     **osqp_kwargs,
 ) -> Array:
     r"""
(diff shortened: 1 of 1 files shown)


coreax/util.py
```diff
@@ -239,7 +246,10 @@ def solve_qp(kernel_mm: ArrayLike, gramian_row_mean: ArrayLike, **osqp_kwargs) -
     sol = qp.run(
         params_obj=(q_array, c), params_eq=(a_array, b), params_ineq=(g_array, h)
     ).params
-    return sol.primal
+
+    # Ensure conditions of solution are met to chosen precision
+    solution = jnp.where(sol.primal < jnp.abs(precision_threshold), 0, sol.primal)
```
pc532627, line 251: I think it might be worth adding a unittest to verify the threshold is applied correctly. Something like, mock/patch the qp.run, forcing it to return some values above the precision threshold and some values below, then verify that the ones below have been filled in.

pc532627 (changes requested):
One small unittest request but otherwise looks good.

coreax/util.py
db091756: I forgot that we already have a tested `apply_negative_precision_threshold` function so I have now used that instead

pc532627 (approved):

PR #708, opened 2024-07-23
refactor kernels.py into submodule

db091756:
### PR Type

### Description

Creates `kernels` subpackage.

`kernels.util.py` holds utility functions relating to kernels.
`kernels.base.py` holds the base kernel class `ScalarValuedKernel`, as well as the base composite classes `UniCompositeKernel`,  `DuoCompositeKernel` and the derived duo-composite kernels `AdditiveKernel` and `ProductKernel`
`kernels.scalar_valued.py` holds all currently implemented scalar valued kernels.

### How Has This Been Tested?

All existing tests pass.

(Write your answer here.)

### Screenshots

(Write your answer here.)


coreax/kernels/base.py
```diff
@@ -0,0 +1,503 @@
...
+
+
+class ScalarValuedKernel(eqx.Module):
+    """Abstract base class for scalar-valued kernels."""
+
+    def __add__(self, addition: ScalarValuedKernel) -> AdditiveKernel:
+        """Overload `+` operator."""
+        if not isinstance(addition, ScalarValuedKernel):
+            raise ValueError("'addition' must be an instance of a 'ScalarValuedKernel'")
+        return AdditiveKernel(self, addition)
```
rg936672, line 78: We've lost the ability to directly add a scalar to a kernel. (Also, if that's added back in, we'll need to reimplement `__radd__` to allow `kernel + scalar` as well as `scalar + kernel`).

coreax/kernels/base.py
```diff
@@ -0,0 +1,503 @@
...
+
+    def __mul__(self, product: ScalarValuedKernel) -> ProductKernel:
+        """Overload `*` operator."""
+        if not isinstance(product, ScalarValuedKernel):
+            raise ValueError("'product' must be an instance of a 'ScalarValuedKernel'")
+        return ProductKernel(self, product)
```
rg936672, line 84: We've also lost the ability to directly multiply a kernel by a scalar. (And we'll need to reimplement `__rmul__` to make it work both ways).

coreax/kernels/base.py
```diff
@@ -0,0 +1,503 @@
...
+
+        # Ensure the first and second kernels are symmetric for even powers to make use
+        # of reduced computation capabilities in ProductKernel. For example,
+        # :meth:`~divergence_x_grad_y_elementwise` can be computed more efficiently if
+        # the first and second kernels are recognised as the same.
+        for i in range(min_power, power):
+            if i % 2 == 0:
+                first_kernel = ProductKernel(first_kernel, self)
+            else:
+                second_kernel = ProductKernel(second_kernel, self)
```
rg936672, line 104: Would it be more efficient to instead implement a `PowerKernel` subclassing `UniCompositeKernel` that computes this directly, rather than relying on nested `ProductKernel` instances? That would also allow non-integer powers. That's probably out-of-scope for this PR, though, so may want to be its own issue.

coreax/kernels/base.py
```diff
@@ -0,0 +1,503 @@
...
+            block_sum, jnp.asarray(0.0), block_x, unroll=outer_unroll
+        )
+        if axis is None:
+            return accumulated_sum
+        num_rows_padded = block_x.data.shape[0] * block_x.data.shape[1]
+        column_sum_padded = column_sum_blocks.reshape(num_rows_padded, -1).sum(axis=1)
+        return column_sum_padded[:unpadded_len_x]
+
+
+class UniCompositeKernel(ScalarValuedKernel):
```
rg936672, line 350: Would it make more sense for this (and `DuoCompositeKernel`, `AdditiveKernel`, `ProductKernel`) to be in `coreax.kernels.scalar_valued` instead? Presumably when vector/function-valued kernels are introduced, we'll want to also have an `AdditiveKernel` (etc.) equivalent for them as well.

CHANGELOG.md
```diff
@@ -20,6 +20,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `coreax.inverses.RegularisedInverseApproximator` replaced by `coreax.least_squares.RegularisedLeastSquaresSolver`
   - `coreax.inverses.LeastSquaresApproximator` replaced by `coreax.least_squares.MinimalEuclideanNormSolver`
   - `coreax.inverses.RandomisedEigendecompositionApproximator` replaced by `coreax.least_squares.RandomisedEigendecompositionSolver`
+- Refactoring of `coreax.kernel.py` into `coreax.kernels` sub-package:
+  - `kernels.util.py` holds utility functions relating to kernels e.g. `median_heuristic`.
```
rg936672, line 24: Mention that `block_data_convert` is now public.

coreax/kernels/base.py
rg936672: Note that this (and the change to multiplication below) is a breaking change. Thus, it should have a deprecation period. (Or just not be changed - this feels like an unnecessary feature removal!)

coreax/kernels/base.py
rg936672: (If moved, remember to change the changelog comment as well!)

rg936672 (changes requested):
Generally looks good, but several changes here are breaking changes, which will need to have deprecation notices.

coreax/kernels/base.py
```diff
@@ -0,0 +1,503 @@
...
+    def __mul__(self, product: ScalarValuedKernel) -> ProductKernel:
+        """Overload `*` operator."""
+        if not isinstance(product, ScalarValuedKernel):
+            raise ValueError("'product' must be an instance of a 'ScalarValuedKernel'")
+        return ProductKernel(self, product)
+
+    def __pow__(self, power: int) -> ProductKernel:
+        """Overload `**` operator."""
+        min_power = 2
+        power = int(power)
```
rg936672, line 89: Note that doing `int(power)` just rounds `power` to the nearest integer - it won't throw any kind of error. To see this, change the test for invalid inputs to `__pow__` to instead to `kernel ** 2.5`, and note that no error is thrown (as the float is silently rounded to an int).

coreax/kernels/base.py
db091756: See above comment

coreax/kernels/base.py
db091756: Yes that seems like a good idea, will try implement this

coreax/kernels/base.py
db091756: If we move these classes then I have come into issues with circular imports as we have to import from the `scalar_valued.py` file which itself imports from `base.py` already. 

I couldn't come up with a way to have `AdditiveKernel` for example live outside the same class as `ScalarValuedKernel` for example

CHANGELOG.md
db091756: Will make this a private function again

coreax/kernels/base.py
db091756: I did realise this was happening but given that you cannot, in general, take fractional powers of kernels and [expect them to still be valid kernels](https://stats.stackexchange.com/questions/566822/in-machine-learning-for-a-kernel-function-k-is-sqrtk-also-a-valid-kernel-fun) I decided to round the float. I can see how this could be confusing and will make this error in the implementation of `PowerKernel`

rg936672: Ah, that's fair enough, I wasn't aware of that issue. Although just looking at the answer you linked, is the issue there that the exponent is less than 1, rather than that the exponent is fractional? I'm afraid I'm not familiar enough with kernels to answer that myself!

coreax/kernels/base.py
rg936672: You _could_ do 

```python
# __init__.py
from coreax.kernels.base import ScalarValuedKernel
from coreax.kernels.scalar_valued import (AdditiveKernel, ProductKernel, ...)

# base.py
from coreax import kernels
def function(x: kernels.AdditiveKernel): ...

# scalar_valued.py
from coreax import kernels
def other(x: kernels.ScalarValuedKernel): ...
```

and this avoids the circular import issue with some import-logic wizardry.

Alternatively, and I think this is possibly a better solution, you could move `base.py` and `scalar_valued.py` to e.g. `scalar_valued/base.py` and `scalar_valued/kernels.py` respectively, so that in future the other-valued kernels could be in e.g. `function_valued/` and `vector_valued/`. Then when importing from elsewhere, rather than `from coreax.kernels import XYZ`, you instead have to specify and go `from coreax.kernels.scalar_valued import XYZ`. (Then we can separately have `coreax.kernels.scalar_valued.AdditiveKernel` and `coreax.kernels.vector_valued.AdditiveKernel`; if client code needs to use both in the same file they can always just `import coreax.kernels.scalar_valued.AdditiveKernel as ScalarAdditiveKernel`.

coreax/kernels/base.py
rg936672: It's not necessarily a significant loss, but it _is_ a breaking change, so we shouldn't make it without a deprecation period. (In this particular case, we can just move `LinearKernel` back to `base.py`, and it then Just Works™.)

coreax/kernels/base.py
db091756: Implemented this now

coreax/kernels/base.py
db091756: The positive semi-definiteness of kernel functions is very related to the positive semi-definiteness of their resulting (kernel matrices)[https://en.wikipedia.org/wiki/Positive-definite_function]. So if we are considering fractional powers:

Given some kernel function $k:\mathcal{X}\times\mathcal{X}\to\mathbb{R}$ define

$$p(x,y) = k(x,y)^n$$

where

coreax/kernels/base.py
db091756: Added a private `_Constant` class to get around this issue.

db091756: Similar to https://scikit-learn.org/stable/modules/generated/sklearn.gaussian_process.kernels.ConstantKernel.html

coreax/kernels/base.py
db091756: Added back in

coreax/kernels/base.py
db091756: I am having trouble doing `from coreax import kernels` within the `base.py` and `scalar_valued.py` files, it appears to not recognise the import. Perhaps we can revisit this when I finish the operator-valued kernel implementation to see if this is an issue.

CHANGELOG.md
```diff
@@ -20,6 +20,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   - `coreax.inverses.RegularisedInverseApproximator` replaced by `coreax.least_squares.RegularisedLeastSquaresSolver`
   - `coreax.inverses.LeastSquaresApproximator` replaced by `coreax.least_squares.MinimalEuclideanNormSolver`
   - `coreax.inverses.RandomisedEigendecompositionApproximator` replaced by `coreax.least_squares.RandomisedEigendecompositionSolver`
+- Refactoring of `coreax.kernel.py` into `coreax.kernels` sub-package:
```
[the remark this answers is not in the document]
db091756: Done this now

coreax/kernels/base.py
rg936672: Revisiting this with the vector/function-valued kernels seems sensible, so let's leave this until then.

coreax/kernels/base.py
rg936672: Might be worth mentioning this in the `ScalarValuedKernel` or `ScalarValuedKernel.__pow__` docstring.

coreax/kernels/base.py
```diff
@@ -347,6 +351,37 @@ def slice_sum(accumulated_sum: Array, y_block: Data) -> tuple[Array, Array]:
...
 
 
+class _Constant(ScalarValuedKernel):
+    r"""Define a helper class to add additional functionality to magic methods."""
+
+    constant: float = eqx.field(default=1, converter=float)
+
+    def __check_init__(self):
+        """Check that the constant is not negative."""
+        if self.constant < 0:
```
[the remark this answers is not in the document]
db091756: Yes, added. Also added an additional parameter

coreax/kernels/scalar_valued.py
```diff
@@ -30,28 +30,42 @@ class LinearKernel(ScalarValuedKernel):
...
+    Given :math:`\rho`=` ``output_scale``, :math:`a=` ``constant`` and
+    :math: `b =` ``offset``,  the linear kernel is
     defined as :math:`k: \mathbb{R}^d\times \mathbb{R}^d \to \mathbb{R}`,
-    :math:`k(x, y) = \rho x^Ty + c`.
+    :math:`k(x, y) = a + \rho (x-b)^T(y-b)`.
 
     :param output_scale: Kernel normalisation constant, :math:`\rho`
-    :param constant: Additive constant, :math:`c`
+    :param constant: Additive constant, :math:`a`, must be positive
+    :param offset: Offset term, :math:`b`
```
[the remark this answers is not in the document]
db091756: Will do

coreax/kernels/scalar_valued.py
```diff
@@ -30,28 +30,42 @@ class LinearKernel(ScalarValuedKernel):
...
     """
 
     output_scale: float = eqx.field(default=1.0, converter=float)
     constant: float = eqx.field(default=0.0, converter=float)
+    offset: float = eqx.field(default=0.0, converter=float)
+
+    def __check_init__(self):
+        """Check that the additive constant is not negative."""
+        if self.constant < 0:
```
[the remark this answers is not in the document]
db091756: Yes, this would need adding to every kernel in the codebase though. I guess we have been assuming up to this point people choose sensible parameters for these things.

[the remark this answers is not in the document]
db091756: I could add `__check_init__`'s to every kernel or remove the one I added to stay consistent?

rg936672 (approved):

coreax/kernels/scalar_valued.py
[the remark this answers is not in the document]
db091756: See #719

coreax/kernels/scalar_valued.py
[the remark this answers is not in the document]
db091756: See #718

PR #711, opened 2024-07-29
Feature/performance regression testing

db091756:
### PR Type

- Feature
- Tests

### Description

Introduced function `speed_comparison_test`, main use is as a developer aid to check that the effect on speed of code changes.

### How Has This Been Tested?

Test to sense check the `speed_comparison_test`.
Adds a `test_speed` to `test_solvers.py` which runs the `speed_comparison_test` on each implemented solver and produces a dictionary of formatted results for inspection. Needs code to store it.

(Write your answer here.)

### Screenshots

(Write your answer here.)


tp832944 (changes requested):
Some interim suggestions for improvement.

coreax/util.py
```diff
@@ -302,19 +323,100 @@ def jit_test(
...
+) -> tuple[Sequence[tuple[Array, Array]], dict]:
+    """
+    Compare compilation time and runtime of a list of JIT-able functions.
+
+    :param function_setups: Sequence of instances of :class:`JITCompilableFunction`
+    :param num_runs: Number of times to average function timings over
+    :print_results: If :data:`True`, the results are formatted and printed
+    :param check_hash: If :data:`True` check that the hash of the JITted functions are
+        different to the supplied functions.
+    :return: Mean and standard deviation of compilation time and runtime for each
```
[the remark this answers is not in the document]
db091756: They're 1-dimensional arrays

coreax/util.py
```diff
@@ -282,19 +308,100 @@ def jit_test(
...
+                + f"{_format_number(results[k][1][0].item())}"
+                + f" per run (mean ± std. dev. of {num_runs} runs)"
+            )
+            _logger.info(
+                "Execution time: "
+                + f"{_format_number(results[k][0][1].item())} ± "
+                + f"{_format_number(results[k][1][1].item())}"
+                + f" per run (mean ± std. dev. of {num_runs} runs)"
+            )
+    return results, timings_dict
```
tp832944, line 402: Insert a blank line. Looks a bit tight otherwise.

coreax/util.py
```diff
@@ -282,19 +308,100 @@ def jit_test(
...
+                jit_test(*function_setups[i], check_hash=check_hash)
+            )
+        # Compute the time just spent on compilation
+        timings_dict[f"timings_{i}"] = timings.at[:, 0].set(
+            timings[:, 0] - timings[:, 1]
+        )
+        # Compute summary statistics
+        results.append((timings.mean(axis=0), timings.std(axis=0)))
+
+    if log_results:
```
tp832944, line 387: I'd put the `if` statement inside the `i` loop and ditch the `k` loop.

coreax/util.py
```diff
@@ -282,19 +308,100 @@ def jit_test(
...
+        for j in range(num_runs):
+            timings = timings.at[j, :].set(
+                jit_test(*function_setups[i], check_hash=check_hash)
+            )
+        # Compute the time just spent on compilation
+        timings_dict[f"timings_{i}"] = timings.at[:, 0].set(
+            timings[:, 0] - timings[:, 1]
+        )
+        # Compute summary statistics
+        results.append((timings.mean(axis=0), timings.std(axis=0)))
```
tp832944, line 385: If taking my suggestion of ditching the `k` loop, save mean and SD to explicit variables. Then `results.append(mean, sd)` and logging becomes neater with e.g. `mean[0].item()`.

coreax/util.py
```diff
@@ -282,19 +308,100 @@ def jit_test(
...
+
+    if log_results:
+        for k in range(num_functions):
+            _logger.info("------------------- Function %s -------------------", k + 1)
+            _logger.info(
+                "Compilation time: "
```
tp832944, line 391: `Virgin` or `Uncompiled`? `Compilation time` is really the time stored in `timings_dict`.

coreax/util.py
```diff
@@ -282,19 +308,100 @@ def jit_test(
...
+                + f"{_format_number(results[k][0][0].item())} ± "
+                + f"{_format_number(results[k][1][0].item())}"
+                + f" per run (mean ± std. dev. of {num_runs} runs)"
+            )
+            _logger.info(
+                "Execution time: "
```
tp832944, line 397: `Pre-compiled execution time`? Or just `Second run time`? Technically both are `execution times`.

coreax/util.py
```diff
@@ -282,19 +308,100 @@ def jit_test(
...
+    timings_dict = {}
+    results = []
+    for i in range(num_functions):
+        timings = jnp.zeros((num_runs, 2))
+        for j in range(num_runs):
+            timings = timings.at[j, :].set(
+                jit_test(*function_setups[i], check_hash=check_hash)
+            )
+        # Compute the time just spent on compilation
+        timings_dict[f"timings_{i}"] = timings.at[:, 0].set(
```
tp832944, line 381: No need for a string in the key. I would embrace the power of Python and use `function_setups[i]` itself as the key (or `function_setup` if embracing the power of `enumerate`), or just the integer `i` if you don't like that.

tp832944, line 381: Ditch the `at ... set`. Only want to keep the first column. Use `= timings[:, 0] - timings[:, 1]` instead.

coreax/util.py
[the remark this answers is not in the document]
tp832944: I can see what I missed now. However, you do need to make the structure of the data clearer here:

```
List of tuples (means, standard deviations) for each function containing uncompiled and precompiled execution times as array components; Dictionary with key function and value array of execution time savings for each repeat test of a function
```

I'm sure you can improve on my wording.

Note also my points below in the logging section where I take issue with the terms `compilation time` and `run time`.

tests/unit/test_util.py
```diff
@@ -266,6 +268,53 @@ def _mock(x=1.0, *, a=2.0):
...
+            num_points = a.shape[0]
+            total = 0
+            for i in range(num_points):
+                total += a[i]
+            return total / num_points
+
+        random_vector = jr.normal(jr.key(2_024), shape=(100,))
+
+        num_runs = 10
+        summary_stats, _ = speed_comparison_test(
```
tp832944, line 299: Keep the dictionary (second return) too. Let's do some more testing with it on slow mean:

1. Assert that all the uncompiled times are slower than precompiled.
2. Assert that the difference in means for `summary_stats` is (almost) equal to the mean of the dictionary values. [We can't do a similar thing for SD.]

coreax/util.py
```diff
@@ -302,19 +323,100 @@ def jit_test(
...
+
+    return f"{round(scaled_num, 2)} {order}"
+
+
+def speed_comparison_test(
+    function_setups: Sequence[JITCompilableFunction],
+    num_runs: int = 10,
+    print_results: bool = False,
+    check_hash: bool = False,
+) -> tuple[Sequence[tuple[Array, Array]], dict]:
```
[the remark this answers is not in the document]
tp832944: That `float` should be `Array (now that I understand it).

tests/unit/test_solvers.py
```diff
@@ -157,6 +165,32 @@ def test_reduce(
...
+    def test_speed(self, reduce_problem: _ReduceProblem):
+        """Store mean, std. dev. of runs of the compilation and execution time."""
+        dataset, solver, _ = reduce_problem
+        reduction_method = JITCompilableFunction(
+            solver.reduce, fn_kwargs={"dataset": dataset}
+        )
+        results, _ = speed_comparison_test([reduction_method], NUM_SPEED_TEST_RUNS)
+        results = results[0]
+        # dataset shape, solver args, compilation time, runtime
+        result_dict = {}
```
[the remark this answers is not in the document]
tp832944: No longer relevant to this PR.

tests/unit/test_util.py
```diff
@@ -266,6 +268,53 @@ def _mock(x=1.0, *, a=2.0):
         # `post_time` is upper bounded by `pre_time - wait_time`.
         assert post_time < (pre_time - wait_time)
```
tp832944, line 269: Add an assertion that `post_time > 0`. I raise this because the post times I receive when testing `speed_comparison_test` in the console are all zero. A quick scan of the docstring for `time.time` suggests you may not be able to rely on it for fractions of a second, so you may need to find an alternative timing function here.

tp832944 (changes requested):
@db091756 I've given it a spin and identified a few issues. Sorry to give you such a long list of comments!

coreax/util.py
```diff
@@ -46,6 +52,23 @@ class NotCalculatedError(Exception):
     """Raise when trying to use a variable that has not been calculated yet."""
 
 
+class JITCompilableFunction(NamedTuple):
+    """
```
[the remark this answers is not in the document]
db091756: Made these edits, I am not really sure what these functions do so not sure if I have done a good job with the docstrings!

coreax/util.py
```diff
@@ -282,19 +308,100 @@ def jit_test(
     def _fn(*args, **kwargs):
         return fn(*args, **kwargs)
 
-    assert hash(_fn) != hash(fn), "Cannot guarantee recompilation of `fn`."
+    if check_hash:
+        assert hash(_fn) != hash(fn), "Cannot guarantee recompilation of `fn`."
```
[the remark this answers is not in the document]
db091756: This was in the original code that @tc85324 implemented, didn't want to remove it as I wasn't sure when this would be a problem!

coreax/util.py
```diff
@@ -282,19 +308,100 @@ def jit_test(
...
     block_until_ready(_fn(*fn_args, **fn_kwargs))
     end_time = time.time()
     post_delta = end_time - start_time
+
     return pre_delta, post_delta
 
 
+def _format_number(num: float) -> str:
+    """Standardise the format of the input number."""
+    if num == 0:
```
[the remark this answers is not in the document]
db091756: Made it absolute so negative values can be handled

coreax/util.py
```diff
@@ -282,19 +308,100 @@ def jit_test(
...
+        return "0 s"
+    order = log10(num)
+    if order >= 2:  # noqa: PLR2004
+        scaled_num = num / 60
+        order = "mins"
+    elif 0 <= order < 2:  # noqa: PLR2004
```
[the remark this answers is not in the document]
db091756: This would throw an error as we convert order to a string

coreax/util.py
```diff
@@ -258,18 +281,21 @@ def jit_test(
...
+    check_hash: bool = True,
 ) -> tuple[float, float]:
     """
     Verify JIT performance by comparing timings of a before and after run of a function.
 
     The function is called with supplied arguments twice, and timed for each run. These
     timings are returned in a 2-tuple.
 
-    :param fn: Function callable to test
+    :param fn: JIT-compilable function callable to test
```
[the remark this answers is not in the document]
db091756: I didn't want to touch this function too much, I think we want to rely on using these `NamedTuple`s as little as possible as it just makes it harder for the user to interact with the code in my opinion. It was more of a convenience to make the comparison function parameters more succint.

coreax/util.py
db091756: My understanding of what JAX is doing when you run the function the first time is compiling the function, then dispatching it. Then the second run just dispatches the compiled code. So by taking the difference between the second and first times we get an estimate of compilation time?

coreax/util.py
db091756: The first column would be an estimate of compilation time and the second dispatch time I think

tests/unit/test_util.py
db091756: I am not sure if I am following exactly what these tests show?

coreax/util.py
```diff
@@ -282,19 +308,100 @@ def jit_test(
...
+        order = "mins"
+    elif 0 <= order < 2:  # noqa: PLR2004
+        scaled_num = num
+        order = "s"
+    elif -3 <= order < 0:  # noqa: PLR2004
+        scaled_num = 1e3 * num
+        order = "ms"
+    elif -6 <= order < -3:  # noqa: PLR2004
+        scaled_num = 1e6 * num
+        order = "\u03bcs"
```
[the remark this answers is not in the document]
db091756: This seems to be rejected by the file encoding checker in the pre-commit checks

coreax/util.py
db091756: This is rejected as the `NamedTuple` object is unhashable

documentation/source/conf.py
```diff
@@ -171,7 +177,19 @@
     ("py:class", "jax._src.typing.SupportsDType"),
     # TODO: Remove once no longer supporting Numpy < 2
     # https://github.com/gchq/coreax/issues/674
-    ("py:class", "numpy.bool_"),
+    ("py:class", "equinox._module.Module"),
+    ("py:class", "coreax.coreset._Data"),
+    ("py:class", "coreax.solvers.composite._Data"),
+    ("py:class", "coreax.solvers.composite._Coreset "),
```
tp832944, line 183: Remove spurious space before closing quote.

documentation/source/conf.py
```diff
@@ -171,7 +177,19 @@
...
+    ("py:class", "coreax.solvers.composite._State"),
+    ("py:class", "coreax.solvers.base._Data"),
+    ("py:class", "coreax.solvers.base._State"),
+    ("py:class", "coreax.solvers.base._Coreset"),
+    ("py:class", "coreax.solvers.coresubset._Data"),
+    ("py:class", "coreax.solvers.coresubset._State"),
+    ("py:class", "coreax.solvers.coresubset._Coreset"),
+    ("py:class", "coreax.weights._Data "),
```
tp832944, line 191: Remove spurious space before closing quote.

documentation/source/conf.py
```diff
@@ -171,7 +177,19 @@
...
+    ("py:class", "coreax.solvers.coresubset._SupervisedData "),
```
tp832944, line 192: Remove spurious space before closing quote.

documentation/source/conf.py
```diff
@@ -321,3 +339,25 @@ def create_custom_inv_file(
...
     create_custom_inv_file(tqdm, tqdm_refs)
+
+
+# pylint: disable=unused-argument
+# pylint: disable=unidiomatic-typecheck
+# pylint: disable=protected-access
+
+
+def remove_namedtuple_attrib_docstring(app, what, name, obj, skip, options):
+    """Remove auto documented parameters from all NamedTuple classes."""
```
tp832944, line 350: A better description is:
```
Combine fields and parameters into a single entry for NamedTuple classes.
```

No need to populate rest of docstring.

documentation/source/conf.py
```diff
@@ -321,3 +339,25 @@ def create_custom_inv_file(
...
+    return skip
+
+
+# pylint: enable=protected-access
+# pylint: enable=unused-argument
+# pylint: enable=unidiomatic-typecheck
+
+
+def setup(app):
+    """Attaches `remove_namedtuple_attrib_docstring` handler to event."""
```
tp832944, line 362: A more generic description is:
```
Register custom functions.
```
No need to populate rest of docstring.

coreax/util.py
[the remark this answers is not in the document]
tp832944: I've suggested some wordings.

coreax/util.py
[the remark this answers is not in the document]
tp832944: Forgot that was in there. Keep it as is.

coreax/util.py
tp832944: How about just the integer `i`? If you really want a string, go for `f"function_{i}"`.

coreax/util.py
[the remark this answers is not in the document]
tp832944: I'd insert `number` after `key function` (change in light of not being able to store a NamedTuple).

coreax/util.py
```diff
@@ -282,19 +309,109 @@ def jit_test(
...
     post_delta = end_time - start_time
+
     return pre_delta, post_delta
 
 
+def format_time(num: float) -> str:
+    """
+    Standardise the format of the input time.
+
+    Floats will be converted to a standard format, e.g. 0.4531 -> "453.1 ms"
```
tp832944, line 332: Missing full stop.

coreax/util.py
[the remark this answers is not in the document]
tp832944: It shouldn't do. If it does (and maybe you should do this anyway), you're repurposing `order`, and I would define a new variable `unit` to contain the units string.

tests/unit/test_util.py
```diff
@@ -219,13 +222,20 @@ def test_sample_batch_indices_inter_row_uniqueness(self) -> None:
...
             num_batches=num_batches,
         )
         assert jnp.unique(batch_indices, axis=0).shape[0] == num_batches
 
+    def test_format_time(self) -> None:
+        """Test that `format_time` outputs expected strings."""
+        assert format_time(0.0004531) == "453.1 \u03bcs"
+        assert format_time(-0.032) == "-32.0 ms"
+        assert format_time(125) == "2.08 mins"
+        assert format_time(1e-15) == "0.0 ps"
```
tp832944, line 237: Good work incorporating an edge case. However, I would include at least one example for each unit, i.e. I'd like to see `0 s`, `1 s` and `1 ns` examples too.

coreax/util.py
tp832944: Needs to be `mean[0].item()` not `mean[0][0].item()` etc.

coreax/util.py
tp832944: Let's talk.

tests/unit/test_util.py
tp832944: Might become clearer as we talk.

tests/unit/test_util.py
```diff
@@ -246,7 +256,10 @@ def test_jit_test(self, args, kwargs, jit_kwargs) -> None:
         wait_time = 2
         trace_counter = Mock()
 
-        def _mock(x=1.0, *, a=2.0):
+        x = jnp.ones(1000)
```
tp832944, line 259: Give these two variables different names to the parameters in `_mock`, e.g. rename this to `x_in`.

tp832944 (changes requested):

tp832944:
@db091756 There are some pre-commit failures. Merging main will deal with some of the doc build failures.

coreax/util.py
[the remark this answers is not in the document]
tp832944: @db091756 Nudge. You see to have missed this.

coreax/util.py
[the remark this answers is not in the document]
tp832944: Sorry @db091756, I've made a mistake here. Your refactor makes it less efficient - we do want to keep `elif` statements. I would:

- Ditch the definition of `scaled_time` and `unit_string` at the top of the function.
- Catching the exception on `log10` is a good robustness improvement - nice idea! But just return a hardcoded `"0 s"`. It was neater before.
- Use a sequence of `if`, `elif` and `else` statements, defining `scaled_time` and `unit_string` inside as you currently do.
  - Move the `s` into the else.
  - The first `elif` should be for `order < -9`, then `< -6`, finishing with `< -3`.

Or just use this function body after docstring:

``` python
    try:
        order = log10(abs(num))
    except ValueError:
        return f"0 s"

    if order >= 2:  # noqa: PLR2004
        scaled_time = num / 60
        unit_string = "mins"
    elif order < -9:  # noqa: PLR2004
        scaled_time = 1e12 * num
        unit_string = "ps"
    elif order < -6:  # noqa: PLR2004
        scaled_time = 1e9 * num
        unit_string = "ns"
    elif order < -3:  # noqa: PLR2004
        scaled_time = 1e6 * num
        unit_string = "\u03bcs"
    elif order < 0:  # noqa: PLR2004
        scaled_time = 1e3 * num
        unit_string = "ms"
    else:
        scaled_time = num
        unit_string = "s"

    return f"{round(scaled_time, 2)} {unit_string}"
```

coreax/util.py
tp832944: We now agree here. Please update the `:return:` of the docstring to match.

coreax/util.py
```diff
@@ -282,19 +309,110 @@ def jit_test(
...
+    :param function_setups: Sequence of instances of :class:`JITCompilableFunction`
+    :param num_runs: Number of times to average function timings over
+    :param log_results: If :data:`True`, the results are formatted and logged
+    :param check_hash: If :data:`True`, check that the hash of the JITted functions are
+        different to the supplied functions
+    :return: List of tuples (means, standard deviations) for each function containing
+        JIT compilation and execution times as array components; Dictionary with
+        key function number and value array of execution time savings for each repeat
+        test of a function
+
```
tp832944, line 383: Remove spurious blank line.

tp832944 (changes requested):
Nearly there @db091756. Sorry for the mistakes I've made along the way. Scroll up to check for any unresolved conversations. Good news: `perf_counter` now means that I get a non-zero execution time on Windows!

coreax/util.py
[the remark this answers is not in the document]
db091756: Whoops, fixed

tp832944 (approved):

PR #713, opened 2024-07-30
fix: correct problem definitions in `test_weights`

tc85324:
### PR Type
- Bugfix
- Tests

### Description
Coreset indices were not being generated as expected; instead of a set of `self.coreset_size` indices being generated, only a single index in the range of `self.coreset_size` was generated.

This would be fine, except for `test_solver_reduces_target_metric` only being well defined for coreset sizes greater than one. When the coreset size is one, there is no guarantee that solving for the weights will reduce the target metric; the weighting is usually a relative weight against other members of the coreset.

### How Has This Been Tested?
Test pass; manual check of coreset generation in the problem now demonstrates the expected behaviour.

No


db091756 (approved):
Good spot!

PR #738, opened 2024-08-16
Performance testing CI workflow

rg936672:
### PR Type

- CI related changes

### Description

Add a CI workflow for performance testing: 

 * On each push to `main`, it records a set of performance metrics.
 * On each push to a PR, it computes performance metrics and compares them the most recent recorded set of performance metrics from `main`, and then adds a comment to the PR with a summary of any significant differences.

Also makes a couple of changes to `jit_test`:
 * Adds an optional normalisation factor, which is used to reduce variance from running the performance tests across different machines.
 * `JITCompilableFunction`s now have an optional string name (and are named "Function 1", "Function 2", etc if `None` is given).

### How Has This Been Tested?

See usage at https://github.com/rg936672/coreax/pull/1 and https://gist.github.com/rg936672/094213d70dcc9c03ec49a81545627373.

No.

### Screenshots

(Write your answer here.)

### DIFF
--- a/coreax/util.py
+++ b/coreax/util.py
@@ -27,7 +27,7 @@ import time
 from collections.abc import Callable, Iterable, Iterator
 from functools import partial, wraps
 from math import log10
-from typing import Any, NamedTuple, Optional, Sequence, TypeVar
+from typing import Any, NamedTuple, Optional, Sequence, Tuple, TypeVar
 
 import equinox as eqx
 import jax.numpy as jnp
@@ -358,6 +358,7 @@ def speed_comparison_test(
     num_runs: int = 10,
     log_results: bool = False,
     check_hash: bool = False,
+    normalisation: Optional[Tuple[float, float]] = None,
 ) -> tuple[list[tuple[Array, Array]], dict[str, Array]]:
     """
     Compare compilation time and runtime of a list of JIT-able functions.
@@ -367,6 +368,9 @@ def speed_comparison_test(
     :log_results: If :data:`True`, the results are formatted and logged
     :param check_hash: If :data:`True` check that the hash of the JITted functions are
         different to the supplied functions.
+    :param normalisation: Tuple (compilation normalisation, execution normalisation).
+        If provided, returned compilation/execution times are normalised so that this
+        time is 1 time unit.
     :return: Mean and standard deviation of compilation time and runtime for each
         function as a list of tuples, and a dictionary of raw timings
     """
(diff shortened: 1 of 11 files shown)


rg936672:
Things we'll want to add to this in future:
 - More performance tests
 - Recording performance data from PRs, so that performance can be compared both against the most recent `main` and against the most recent known data for the PR

coreax/util.py
```diff
@@ -281,6 +282,7 @@ def jit_test(
     fn_args: tuple = (),
     fn_kwargs: Optional[dict] = None,
     jit_kwargs: Optional[dict] = None,
+    _name: Optional[str] = None,  # UNUSED
```
db091756, line 285: Rather than noting this as unused, could we just do `del _name` and include a comment explaining why this parameter is necessary here

tests/performance/__init__.py
```diff
@@ -0,0 +1 @@
+"""Module containing tests for performance (i.e. speed) of the package."""
```
db091756, line 1: This and other new files needs copyright message

tests/performance/cases/basic_coresets.py
```diff
@@ -0,0 +1,107 @@
...
+    KernelDensityMatching,
+    PCIMQKernel,
+    SquaredExponentialKernel,
+    SteinKernel,
+)
+from coreax.kernels import median_heuristic
+from coreax.solvers import KernelHerding, RPCholesky, SteinThinning
+from coreax.util import JITCompilableFunction
+
+CORESET_SIZE = 100
```
db091756, line 25: I think I would be in favour of putting `num_data_points` and `num_features` as global constants here as well, as this will have an effect on runtime

tests/performance/compare.py
```diff
@@ -0,0 +1,231 @@
...
+    parser.add_argument(
+        "reference_directory",
+        help="The directory containing historic performance data.",
+    )
+    args = parser.parse_args()
+    return Path(args.performance_file), Path(args.reference_directory)
+
+
+def date_from_filename(path: Path) -> Optional[Tuple[datetime.datetime, str]]:
+    """Extract the date from a performance data file name."""
```
db091756, line 37: Would be useful to have params and return in the docstring here

tests/performance/compare.py
```diff
@@ -0,0 +1,231 @@
...
+        hour=int(hour),
+        minute=int(minute),
+        second=int(second),
+        tzinfo=datetime.timezone.utc,
+    ), git_hash
+
+
+def get_most_recent_historic_data(
+    reference_directory: Path,
+) -> Dict[str, Dict[str, float]]:
```
db091756, line 57: Would be useful to have params and return in the docstring here

tests/performance/compare.py
```diff
@@ -0,0 +1,231 @@
...
+        return {}
+
+    most_recent_file = max(files.keys(), key=files.get)
+
+    with open(most_recent_file, "r", encoding="utf8") as f:
+        return json.load(f)
+
+
+def format_run_time(times: Dict[str, float]) -> str:
+    """Format the performance data for a specific test in a human-readable form."""
```
db091756, line 76: Would be useful to have params and return in the docstring here

tests/performance/compare.py
```diff
@@ -0,0 +1,231 @@
...
+
+    :Example:
+        >>> relative_change(2, 3)  # 3 is 50% bigger than 2
+        0.5
+        >>> relative_change(4, 3)  # 3 is 25% less then 4
+        -0.25
+        >>> relative_change(1, 0)  # 0 is 100% less than 1
+        -1
+        >>> relative_change(0, 1)  # 1 is infinitely larger than 0
+        inf
```
db091756, line 137: Would be useful to have params and return in the docstring here

tests/performance/compare.py
```diff
@@ -0,0 +1,231 @@
...
+    try:
+        return change / before
+    except ZeroDivisionError:
+        return float("inf")
+
+
+def get_significant_differences(
+    current_performance: Dict[str, Dict[str, float]],
+    historic_performance: Dict[str, Dict[str, float]],
+) -> List[Tuple[str, List[str]]]:
```
db091756, line 152: Would be useful to have params and return in the docstring here

db091756 (changes requested):
I am happy with the code. But I think it would help future readers of the code to have more explanation in the docstrings. 

I also think it would be a nice QoL feature to include the `units` in the output. I.e. tell us somewhere the length of time each unit actually is. This way we can figure out how long it took to compute the coresets if we wanted to

rg936672:
> I also think it would be a nice QoL feature to include the `units` in the output. I.e. tell us somewhere the length of time each unit actually is. This way we can figure out how long it took to compute the coresets if we wanted to

This will require a bit of finagling, as well as for more data to be stored in the gist. I'll give it a go though, as I agree it would be nice!

db091756:
> > I also think it would be a nice QoL feature to include the `units` in the output. I.e. tell us somewhere the length of time each unit actually is. This way we can figure out how long it took to compute the coresets if we wanted to
> 
> This will require a bit of finagling, as well as for more data to be stored in the gist. I'll give it a go though, as I agree it would be nice!

I am happy to review this without this extra feature, and add it as a new issue

rg936672:
> I am happy to review this without this extra feature, and add it as a new issue

As this will require changing the format of stored data, I'd prefer to get this right first time rather than splitting this!

rg936672:
No clue why pre-commit is failing here; I can't reproduce the issue locally.

db091756:
@rg936672 Is it possible to have a shared normalisation across the compilation and execution time?

rg936672:
> @rg936672 Is it possible to have a shared normalisation across the compilation and execution time? As it is it may be slightly confusing as they both refer to the same `units`

It's certainly _possible_, but I don't think it'd be a good idea - execution is going to be very parallelisable and thus the time will be highly dependent on the number of cores available, whereas I would assume that compilation will be much less so, so depending on the hardware the two might vary independently. 

I can easily change the language on the comment though - how about just referring to them as "compilation units" and "execution units", or even "c-units" and "e-units"?

db091756:
> > @rg936672 Is it possible to have a shared normalisation across the compilation and execution time? As it is it may be slightly confusing as they both refer to the same `units`
> 
> It's certainly _possible_, but I don't think it'd be a good idea - execution is going to be very parallelisable and thus the time will be highly dependent on the number of cores available, whereas I would assume that compilation will be much less so, so depending on the hardware the two might vary independently.
> 
> I can easily change the language on the comment though - how about just referring to them as "compilation units" and "execution units", or even "c-units" and "e-units"?

I would be happy with either, up to you!

db091756 (approved):

Issue #764, opened 2024-09-03
Track down `score_matching.py` slowdown issue


db091756:
### What's the new feature?

Changing how enforce 2-dimensionality in the `score_matching.py` code slows down `SteinThinning` very significantly. Track down this bug. (See https://github.com/gchq/coreax/pull/746#issuecomment-2321449604)

### What value does this add?

Consistent dimensionality handling without slowdown!

db091756:
> Was this resolved when #746 was merged?

We reverted the line of code that caused the issue. It is not resolved.

PR #775, opened 2024-09-17
docs: Clarify usage of atleast_2d from JAX inside of score matching.

pc532627:
Refs: #764

### PR Type

- Bugfix

### Description

Investigation into score matching slowdown has shown that using both JAX's `jnp.atleast_2d` and `coreax.data._atleast_2d_consistent` are consistent with desired results, but the former is faster when called a huge number of times (as is the case when one might want to use a Stein kernel). A comment has been added explaining this.

### How Has This Been Tested?

N/A

N/A

### DIFF
--- a/coreax/score_matching.py
+++ b/coreax/score_matching.py
@@ -412,7 +412,13 @@ class SlicedScoreMatching(ScoreMatching):
         :param x: The :math:`n \times d` data vectors
         :return: A function that applies the learned score function to input ``x``
         """
-        # Format inputs
+        # Check format of input array. We use atleast_2d from JAX to perform
+        # conversions here. This is faster than using the custom function
+        # _atleast_2d_consistent in coreax.data, and whilst these two functions
+        # differ in output for 1-dimensional arrays, the computations inside of this
+        # function will match even in this case. This speed difference is noticeable
+        # when we use KernelDensityMatching in conjunction with a Stein kernel,
+        # where the score function may be called a huge number of times.
         x = jnp.atleast_2d(x)
 
         # Setup neural network that will approximate the score function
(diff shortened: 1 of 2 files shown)


coreax/score_matching.py
```diff
@@ -545,7 +551,13 @@ def score_function(x_: Shaped[Array, " n d"]) -> Shaped[Array, " n d"]:
             :param x_: The :math:`n \times d` data vectors we wish to evaluate the score
                 function at
             """
-            # Check format
+            # Check format of input array. We use atleast_2d from JAX to perform
+            # conversions here. This is faster than using the custom function
+            # _atleast_2d_consistent in coreax.data, and whilst these two functions
+            # differ in output for 1-dimensional arrays, the computations inside of this
+            # function will match even in this case. This speed difference is noticeable
```
rg936672, line 558: > the computations inside this function will match

This isn't actually correct - running
```python
import jax.numpy as jnp

from coreax import KernelDensityMatching

matcher = KernelDensityMatching(length_scale=1)
data = jnp.array([[1, 2], [3, 4], [5, 6], [7, 8]])

fn = matcher.match(data)
print(fn(jnp.array([1, 2])))
```

with the current code (using `jnp.atleast_2d`) gives `[0.03597286 0.03597286]`. However, if you change line 562 to use `_atleast_2d_consistent` then this example gives `[0.00494525 1.0049453]` instead.

Looking at the type hints involved, and also at line 576 below, I think that `jnp.atleast_2d` is still the correct way to go, but I don't think that this justification is correct.

coreax/score_matching.py
```diff
@@ -545,7 +551,13 @@ def score_function(x_: Shaped[Array, " n d"]) -> Shaped[Array, " n d"]:
             :param x_: The :math:`n \times d` data vectors we wish to evaluate the score
                 function at
             """
-            # Check format
+            # Check format of input array. We use atleast_2d from JAX to perform
+            # conversions here. This is faster than using the custom function
+            # _atleast_2d_consistent in coreax.data, and whilst these two functions
```
rg936672, line 556: > This is faster than using the custom function `_atleast_2d_consistent`

I'm not sure this is true:

```
> python -m timeit --setup "import jax.numpy as jnp; from coreax.data import _atleast_2d_consistent; arr = jnp.arange(100)" -- "_atleast_2d_consistent(arr)"
50000 loops, best of 5: 5.14 usec per loop
> python -m timeit --setup "import jax.numpy as jnp; from coreax.data import _atleast_2d_consistent; arr = jnp.arange(100)" -- "jnp.atleast_2d(arr)"
50000 loops, best of 5: 5.67 usec per loop
```
so the two seem to have nearly identical running times.

The issue instead seems to be that because the output shape is different, we have to do more processing on lines 565 and 568 (as their output shape and output values become different.)

rg936672 (commented):
I think the conclusion to stick with `jnp.atleast_2d` over `_atleast_2d_consistent` is correct, but I'm not sure your reason for why is correct!

rg936672 (approved):
This looks good! Might want to merge #783 first and then re-merge `main` into this branch to force the pre-commit to run on the latest version.

PR #780, opened 2024-09-20
Feature/replace require ascii with utf8

### LINKED ISSUE #768: Replace require-ascii check with require-utf8 check
### What's the issue?

This should be a simple change; just delete the lines from require-ascii that check that each character is in the ASCII range.

This was brought up as the performance monitoring code wants to use the `±` character, which is not in the ASCII range.

rg936672:
### PR Type

- CI related changes

### Description

Closes #768. This PR replaces the require-ascii hook with a require-utf8 hook. Instead of requiring all text files are ASCII, it instead requires that they are valid UTF-8.

### How Has This Been Tested?

Pre-commit passes. Introducing an invalid file (e.g. an image file) to the repo causes it to fail.

No (CI changes only.)


pc532627 (approved):
Changes look good - if the workflows finish and pass I'll merge into develop

Issue #779, opened 2024-09-19
Fix MapReduce solver Indexing bug


qh681248:
### What's the problem?

Mapreduce does not compute indices correctly

If you add plt.show() and run examples/pounce_map_reduce.py, you will see that only the first 10 frames are being chosen from the 106 frames, this is because coreset.unweighted_indices is not working properly

The example worked as of commit d99d4e4 and ran faster

### How can we reproduce the issue?

Run examples/pounce_map_reduce.py and either print the indices of the coreset generated by map reduce or add plt.show() at the end

### Python version

3.12

### Package version

0.2.1

### Operating system

Windows

Issue #765, opened 2024-09-03
Look into typing `Data` and `SupervisedData` instances


db091756:
### What's the new feature?

I think it would be very useful for developers and users to be able to type the array shapes in the `Data` and `SupervisedData` instances.

For example, if we want to type a function input we would do  `x: Data` or `x: SupervisedData`. I think it would be nice to be able to type this something like `x: Data[Shaped[Array, " n d"], Shaped[Array, " n"]]` and `x: SupervisedData[Shaped[Array, " n d"], Shaped[Array, " n p"], Shaped[Array, " n"]]` , or just `x: Data[" n d", "n"]` and `x: SupervisedData[" n d", " n p", "n"]` . 

I have no idea how much work this would be!

### What value does this add?

Would make it clearer what our data instances contain.

PR #790, opened 2024-09-25
Map reduce index bug

qh681248:
### PR Type

- Bugfix

### Description

In coreax/solvers/composite.py, added logic to keep track of indices in the reduce method of MapReduce class, this resolves the Fix MapReduce solver Indexing bug #779

### How Has This Been Tested?

Added test_mapreduce_diverse_selection in tests/units/test_solvers.py. It only passes when MapReduce contains datapoints with indices higher than the size of the leaf_node.

No

### Screenshots

(Write your answer here.)


tc85324 (commented):
Thanks for spotting this bug, I should have caught this earlier! I think the issue is more fundamental to how `Coresubset.coreset` operates, and the error can be fixed by the following change (subject to correcting the coreset materialization behaviour).
```python
           # old
            _coreset = jtu.tree_map(jnp.concatenate, coreset_ensemble)
            return _reduce_coreset(_coreset.coreset)
           # proposed
            _coreset = jtu.tree_map(jnp.concatenate, coreset_ensemble.coreset)
            return _reduce_coreset(_coreset)
```
I'd like to have a look today to see if there is a nice change that can be made to the `Coresubset.coreset` materialization behaviour, before the specific changes proposed here are considered.

tc85324 (commented):
Having taken a look yesterday, I can't find a particularly satisfying solution to the materialization behaviour. We could add specific logic for handling "batched" coresets in the coresubset materialization method `Coresubset.coreset`, but this would require us to handle a variety of different behaviours. E.G. what happens when we have unbatched nodes/indices, but a batched `pre_coreset_data`, what happens when the nodes/indices are batched but `pre_coreset_data` is not (this is the specific problem in our case), etc... If there is wider interest in properly handling batched data, then this may be worth revisiting in a separate issue, perhaps at the same time as #765.

In any case, I'm pleased you've identified the problem (which also indicates the existing MapReduce tests are insufficient) and proposed a solution in this PR, which I will be glad to review today.

bk958178 (changes requested):
@qh681248 See requested changes in the comments below

coreax/solvers/composite.py
```diff
@@ -125,18 +125,58 @@ def reduce(
...
-            return _reduce_coreset(_coreset.coreset)
 
-        coreset_wrong_pre_coreset_data, output_solver_state = _reduce_coreset(dataset)
+            if _indices is not None:
+                final_indices = _indices[concatenated_indices]
+            else:
+                final_indices = concatenated_indices
+            return _reduce_coreset(_coreset, final_indices)
+
+        (coreset_wrong_pre_coreset_data, output_solver_state, _indices) = (
```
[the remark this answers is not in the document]
tc85324: I second this, I can't remember why I chose this variable name and it isn't obvious from the code!

coreax/solvers/composite.py
```diff
@@ -183,4 +223,11 @@ def _binary_tree(_input_data: Data) -> np.ndarray:
         return node_indices.reshape(n_leaves, -1).astype(np.int32)
 
     indices = jax.pure_callback(_binary_tree, result_shape, padded_dataset)
-    return dataset[indices]
+    return dataset[indices], jnp.arange(len(dataset))[indices]
```
tc85324, line 224: I think the return value should be `padded_dataset[indices]`. This bug was in the code before this PR and is another indication that the existing tests are insufficient. Please can you modify/add a test so that this behaviour is properly tested.

tc85324 (changes requested):
I've added a few more comments on top of those from @bk958178. Nothing major, but we might need to look at improving the solver tests (the scope of which may exceed this PR).

tp832944:
@qh681248 This is a significant bug fix. Don't forget to add it to the CHANGELOG.

bk958178 (changes requested):
@qh681248 Thanks for addressing my previous comments. I've left a few further suggestions on your latest changes. 

Please update the CHANGELOG.md. Under 'Fixed', add a bulletpoint describing the bug-fix you've addressed.

tests/unit/test_solvers.py
```diff
@@ -850,6 +850,34 @@ def test_base_solver(
...
+        dataset_size = 40
+        data_dim = 5
+        coreset_size = 6
+        leaf_size = 12
+
+        key = jr.PRNGKey(0)
+        dataset = jr.normal(key, shape=(dataset_size, data_dim))
+
+        kernel = SquaredExponentialKernel()
+        base_solver = KernelHerding(coreset_size=coreset_size, kernel=kernel)
```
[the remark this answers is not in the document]
bk958178: @tc85324 For the sake of minimising overhead, and since we both already have visibility on this PR, I’m happy for the changes that rework the existing tests to be made here.

I don’t think it warrants creating a new issue and PR.

tests/unit/test_solvers.py
```diff
@@ -850,6 +850,165 @@ def test_base_solver(
...
+                [forward_indices, backward_indices], axis=1
+            ).ravel()
+
+            if interleaved_base_solver.coreset_size % 2 != 0:
+                interleaved_indices = jnp.append(interleaved_indices, half_size)
+            return Coreset(dataset[interleaved_indices], dataset), solver_state
+
+        interleaved_base_solver.reduce = interleaved_mock_reduce
+
+        original_data = Data(
```
bk958178, line 980: This doesn't need to span lines 980-1004. Also, the expected type is `Array` and this is a list.

Suggested change: 

    original_data = Data(jnp.array([
            10, 20, 30, 210, 40, 60, 180, 90, 150, 70,
            120, 200, 50, 140, 80, 170, 100, 190, 110, 160, 130
            ]))

tests/unit/test_solvers.py
```diff
@@ -850,6 +850,165 @@ def test_base_solver(
...
+                80,
+                170,
+                100,
+                190,
+                110,
+                160,
+                130,
+            ]
+        )
+        expected_coreset_data = Data([0, 110, 90])
```
bk958178, line 1005: Expects type `Array`. Suggested change:

    expected_coreset_data = Data(jnp.array([0, 110, 90]))

tests/unit/test_solvers.py
```diff
@@ -850,6 +850,165 @@ def test_base_solver(
...
+        partitions_represented = jnp.unique(selected_indices // leaf_size)
+        assert (
+            len(partitions_represented) > 1
+        ), "MapReduce should select points from multiple partitions"
+
+    def test_map_reduce_analytic(self):
+        r"""
+        Test ``MapReduce`` on an analytical example, enforcing a unique coreset.
+
+        In this example, We start with the original dataset
```
bk958178, line 884: Remove capital 'w' on 'We', or remove 'In this example,'

tests/unit/test_solvers.py
```diff
@@ -850,6 +850,165 @@ def test_base_solver(
...
+        2. 2 partitions of size :math:`\lceil 10.5 \rceil = 11` each (with one padded 0)
+        3. 4 partitions of size :math:`\lceil 5.25 \rceil = 6` each (with 3 padded 0's)
+        4. 8 partitions of size :math:`\lceil 2.625 \rceil = 3` each (with 3 padded 0's)
+
+        Since we set the maximum leaf size :math:`m = 6`, we choose the largest
+        partition size that is less than or equal to 6. Thus, we have 4 partitions
+        each of size 6.
+
+        We first pad the dataset by adding 3 zeros, then we arrange the
+        data in ascending order (Make sure to ask if this is another bug,
```
bk958178, line 904: This looks like a note you've left to yourself. Did you get an answer to this question?

Either way, please delete from the docstring

tests/unit/test_solvers.py
```diff
@@ -850,6 +850,165 @@ def test_base_solver(
...
+        I don't know how exactly binary_tree manages to reorder my dataset every time)
```
bk958178, line 905: See above comment - please delete this note you've left for yourself

tests/unit/test_solvers.py
```diff
@@ -850,6 +850,165 @@ def test_base_solver(
...
+        This results in
```
bk958178, line 906: combine this with the following line:

            This results in the following 4 partitions (see how data is in ascending order):

tests/unit/test_solvers.py
```diff
@@ -850,6 +850,165 @@ def test_base_solver(
...
+        the following 4 partitions (see how data is in ascending order):
+
+        1. :math:`[0, 0, 0, 10, 20, 30]`
+        2. :math:`[40, 50, 60, 70, 80, 90]`
+        3. :math:`[100, 110, 120, 130, 140, 150]`
+        4. :math:`[160, 170, 180, 190, 200, 210]`
+
+        Now we want to reduce each partition with our ``interleaved_base_solver``
+        which is designed to choose first, last, second, second-last, third,
```
bk958178, line 915: Suggested rephrasing for clarity:

' is designed to choose 2n elements, with indices [1, 2, ..., n, -n, ..., -2, -1]'

Happy for you to ignore this suggestion if you think it's less clear than the one you've provided

tests/unit/test_solvers.py
```diff
@@ -850,6 +850,165 @@ def test_base_solver(
...
+
+        1. :math:`[0, 30, 0]`
+        2. :math:`[40, 90, 50]`
+        3. :math:`[100, 150, 110]`
+        4. :math:`[160, 210, 170]`
+
+        Concatenating we obtain
+        :math:`[0, 30, 0, 40, 90, 50, 100, 150, 110, 160, 210, 170]`.
+        Now we repeat the same process, we check how many partitions
+        (has to be power of 2) we want to divide this new data of size 12 into,
```
bk958178, line 927: Grammar needs correcting. Suggested rephrasing:

`We repeat the process, checking how many partitions we want to divide this intermediate dataset (of size 12) into. Recall, this number of partitions must be a power of 2. Our options are:`

tests/unit/test_solvers.py
```diff
@@ -850,6 +850,165 @@ def test_base_solver(
...
+
+        1. :math:`[0, 90, 0]`
+        2. :math:`[100, 210, 110]`
+
+        Now, we concatenate the two subsets and repeat the process to
+        obtain only one partition:
+
+        1. Concatenated subset: :math:`[0, 90, 0, 100, 210, 110]`
+
+        Note that the size of the dataset is 6 so no partitioning is necessary!
```
bk958178, line 954: `no *more* partitioning is necessary`

tests/unit/test_solvers.py
```diff
@@ -850,6 +850,165 @@ def test_base_solver(
...
+
+        Applying the ``interleaved_base_solver`` one last time we obtain
```
bk958178, line 956: Suggested rephrasing for conciseness:

    Applying ``interleaved_base_solver`` one last time we obtain the final coreset:
         :math:`[0, 110, 90]`.

bk958178 (changes requested):
Thanks for adding this analytic test, @qh681248. Your description is very thorough. Please see some initial comments and suggested changes below. 

I'll allow time for @tc85324 to review the analytic test as well, and share any thoughts or comments too.

bk958178:
> Thanks for adding this analytic test, @qh681248. Your description is very thorough. Please see some initial comments and suggested changes below.
> 
> I'll allow time for @tc85324 to review the analytic test as well, and share any thoughts or comments too.

@tc85324 if we haven’t heard from you by Thursday, we’ll assume you’re okay with not reviewing these changes before the v0.3 release. Any changes you request later will be addressed and included in future releases.

tests/unit/test_solvers.py
```diff
@@ -923,9 +920,10 @@ def test_map_reduce_analytic(self):
 
         Concatenating we obtain
         :math:`[0, 30, 0, 40, 90, 50, 100, 150, 110, 160, 210, 170]`.
-        Now we repeat the same process, we check how many partitions
-        (has to be power of 2) we want to divide this new data of size 12 into,
-        our new options for partitioning are:
+        We repeat the process, checking how many partitions we want
+         to divide this intermediate dataset (of size 12) into.
+        Recall, this number of partitions must be a power of 2.
+        Our options are::
```
bk958178, line 926: Reduce to 3 lines, and delete duplicated colon

    We repeat the process, checking how many partitions we want to divide this
    intermediate dataset (of size 12) into. Recall, this number of partitions must
    be a power of 2. Our options are:

tests/unit/test_solvers.py
```diff
@@ -850,6 +850,166 @@ def test_base_solver(
...
+        obtain only one partition:
+
+        1. Concatenated subset: :math:`[0, 90, 0, 100, 210, 110]`
+
+        Note that the size of the dataset is 6,
+        therefore, no more partitioning is necessary.
+
+        Applying the ``interleaved_base_solver`` one last time we obtain
+         the final coreset:
+                :math:`[0, 110, 90]`.
```
bk958178, line 957: Reduce to 2 lines:

     Applying ``interleaved_base_solver`` one last time, we obtain the final coreset:
                :math:`[0, 110, 90]`.

tests/unit/test_solvers.py
bk958178:
```suggestion
        original_data = Data(jnp.array([10, 20, 30, 210, 40, 60, 180, 90, 150, 70, 120,
                                        200, 50, 140, 80, 170, 100, 190, 110, 160, 130])
                             )
```

bk958178 (changes requested):
@qh681248 Please see below a few further comments to reduce the number of new lines added

tc85324 (commented):
> > Thanks for adding this analytic test, @qh681248. Your description is very thorough. Please see some initial comments and suggested changes below.
> > 
> > I'll allow time for @tc85324 to review the analytic test as well, and share any thoughts or comments too.
> 
> @tc85324 if we haven’t heard from you by Thursday, we’ll assume you’re okay with not reviewing these changes before the v0.3 release. Any changes you request later will be addressed and included in future releases. 

Unfortunately I don't have the time right now to review this properly. I did have a quick look though and the new tests appear to do what I was hoping for, thanks @qh681248!

bk958178 (approved):
LGTM

Approving and merging

PR #798, opened 2024-10-02
feat: added cubature recombination example.

### LINKED ISSUE #778: Write integration test for tree recombination
### What's the issue?

No end-to-end test exists for tree recombination, which also means we're lacking an example.

tc85324:
Closes #778

### PR Type
- Feature
- Tests

### Description
Adds a test to demonstrate the end-to-end usage of `TreeRecombination`
for constructing non-product cubature formulae from product cubature
formulae. A corresponding integration test has also been added for this
new example.

### How Has This Been Tested?
The new tests pass locally and contain theoretically grounded assertions.


examples/cubature_recombination.py
```diff
@@ -0,0 +1,194 @@
...
+    raise ValueError("Invalid mode; must be one of ['all', 'even', 'odd'].")
+
+
+def test_functions_from_monomial_powers(
+    monomial_powers: Int[Array, "k n"],
+) -> Callable[[Float[Array, " n"]], Float[Array, " k"]]:
+    """
+    Construct test functions given a set of multi-variate monomial powers.
+
+    :param monomial_powers:
```
pc532627, line 108: I think this has not been filled in before committing but should be

examples/cubature_recombination.py
```diff
@@ -0,0 +1,194 @@
...
+    :return:
```
pc532627, line 109: I think this has not been filled in before committing but should be

examples/cubature_recombination.py
```diff
@@ -0,0 +1,194 @@
...
+        return jnp.prod(jnp.polyval(_coefficients[::-1], x))
+
+    return jtu.Partial(monomial_test_function, reversed_coefficients(monomial_powers))
+
+
+def main(dimension: int = 3, max_degree: int = 4) -> Coresubset:
+    """
+    Run the 'recombination' example for finding non-product cubature formulae.
+
+    Generates the
```
pc532627, line 134: I think this has not been filled in before committing but should be

pc532627: Can you add some mathematical background on the problem we are trying to solve here when updating?

pc532627 (changes requested):
A few documentation comments - additionally the doc builds fail - this might be resolved simply by merging in main to this branch

pc532627:
@tc85324 Related to end-to-end usage of recombination, I'd be interested in your thoughts on how to sensibly include Recombination into the benchmarking currently being done on PR #802 . In that PR, @qh681248  is exploring how the various coreset methods compare in terms of run-time and error (MMD, KSD, ...) on a suite of candidate test problems. Since recombination isn't an `ExplicitSizeSolver` direct comparisons to coresets from other methods of a given size may not be in some sense fair. We could exclude it from benchmarking because of this - but if you have any ideas on sensible ways to include it we'd be open to suggestions.

tc85324 (commented):
> @tc85324 Related to end-to-end usage of recombination, I'd be interested in your thoughts on how to sensibly include Recombination into the benchmarking currently being done on PR #802 . In that PR, @qh681248  is exploring how the various coreset methods compare in terms of run-time and error (MMD, KSD, ...) on a suite of candidate test problems. Since recombination isn't an `ExplicitSizeSolver` direct comparisons to coresets from other methods of a given size may not be in some sense fair. We could exclude it from benchmarking because of this - but if you have any ideas on sensible ways to include it we'd be open to suggestions.  

Thank you for the review! Regarding benchmarking, I would say that the example in this pull is probably the canonical use case for recombination and it would be useful to have runtime performance figures for varying numbers of test-functions (dimension and degree), and in the case of `TreeRecombination`, `tree_reduction_factor`. While this won't be comparible to the `ExplicitSizeSolvers` it is useful from an implementaiton perspective as we can check/see if real world performance scales at the same/simlar rates to those claimed in the literature.

pc532627 (approved):

PR #806, opened 2024-10-14
Migrate to uv

### LINKED ISSUE #804: Switch to uv for dependency management
### What's the issue?

uv is a faster, more reproducible package manager than pip.

Replace requirements.txt with uv.lock. Simplifies and speeds up CI.

rg936672:
### PR Type

- Build related changes
- CI related changes

### Description

Migrate to `uv` for dependency management. Closes #804.

### How Has This Been Tested?

See if all the workflows still pass.

Not to end users!


rg936672:
I've hidden it as outdated, but https://github.com/gchq/coreax/pull/806#issuecomment-2410702965 is a very good illustration of just how important having the latest dependencies is for performance - we experience an up-to-10x slowdown with older package versions!

pyproject.toml
```diff
@@ -68,6 +69,14 @@ dev = [
     "ruff",
 ]
 
+# ensure we get the latest package versions available for each Python version
```
tp832944, line 72: Capital letter please.

.github/workflows/docs.yml
```diff
@@ -14,40 +14,40 @@ jobs:
     runs-on: ubuntu-latest
```
tp832944, line 14: Delete ` with latest dependencies` on line above.

tp832944 (changes requested):

.pre-commit-config.yaml
```diff
@@ -53,6 +53,22 @@ repos:
           (?x)^(
             examples/.+/.+.(gif|png)
           )$
+      # Keep requirements-docs is up to date
```
tp832944, line 56: Undo all these changes. This could produce a lot of PRs with an unrelated change when it runs locally on every developer's computer. Let's just rely on the uv autoupdate workflow.

tp832944 (changes requested):
Delete requirements.txt. We have checked that nobody internal requires it.

.pre-commit-config.yaml
rg936672: This will only change requirements-docs if uv.lock changes; as such, it shouldn't make any changes unless the requirements in pyproject.toml change, or unless a developer manually runs `uv lock --upgrade`. (Plus, I'd have to modify the uv autoupdate _again_ to ensure that it updates requirements-docs even if uv.lock doesn't change. I can of course go and make this change, but I think you've possibly misunderstood what the pre-commit hook does here.)

tp832944: Thanks for the explanation.

tp832944 (approved):

PR #821, opened 2024-10-23
:fix: correct plot scaling calculation and leaf parameters

qh681248:
### PR Type

- Bugfix
- Feature

### Description

 The algorithm's execution was slow, not because of image size but due to the leafsize being too close to the coreset_size and there was a bug when we set downsampling factor = 8 which has been fixed here:
- Increase leafsize was from 10,000 to either 16,000, which speeds up the algorithm, even when downsampling = 1.
- Fix scale calculation to use downsampling factor squared
- Update leaf count to 16,000

### How Has This Been Tested?

 Test A: (Write your answer here.)
 Test B: (Write your answer here.)
 Test C: (Write your answer here.)

No

### Screenshots

(Write your answer here.)


gw265981:
From my local testing, I get `RecursionError` before the fix and it runs after the fix (takes around 200s).

examples/david_map_reduce_weighted.py
```diff
@@ -208,7 +227,9 @@ def main(
         -herding_coreset.coreset.data[:, 0],
         c=herding_coreset.coreset.data[:, 2],
         cmap="gray",
-        s=np.exp(2.0 * coreset_size * herding_weights).reshape(1, -1),
+        s=(5.0 * coreset_size * herding_weights * downsampling_factor**2).reshape(
```
rg936672, line 230: This works, but the marker size appears very uniform - this could just be because the weights _are_ quite uniform, though! (They're all in the range `(0.0070, 0.0082)` when I run with `downsampling_factor=8`.) Is it worth fiddling with this to get more of a visual size difference?

examples/david_map_reduce_weighted.py
```diff
@@ -171,6 +189,7 @@ def main(
     # Generate a coreset via uniform random sampling for comparison
     random_solver = RandomSample(coreset_size, sample_key, unique=True)
     random_coreset, _ = eqx.filter_jit(random_solver.reduce)(data)
+    random_weights = weights_optimiser.solve(data, random_coreset.coreset)
```
rg936672, line 192: Should we be computing weights here? Does it make more sense to just set the weight uniformly instead?

rg936672 (changes requested):
Looks good! A couple of questions - if you think these would be better addressed outside this PR I'm happy to approve as is, though.

rg936672 (changes requested):
Sorry to nitpick the changelog entries so much, but I think it's good for us to develop good habits around these! Happy for pushback if you feel I've mischaracterised your changes, though.

CHANGELOG.md
```diff
@@ -19,7 +19,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   points in the coreset.(https://github.com/gchq/coreax/pull/836)
 
 ### Changed
--
+- Adjusted `examples.david_map_reduce_weighted`: increased leaf size to
+  double the coreset size, and set marker sizes of the random and herding
+  coresets to a common scale for fairer comparison.
```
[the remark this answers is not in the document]
qh681248: Sorry, I forgot to link the issue/PR, I have done so now :)

rg936672 (dismissed):
Nice, all looks good :)

rg936672 (approved):

PR #822, opened 2024-10-24
Feature/add analytic test rpcholesky

gw265981:
### PR Type

- Bugfix
- Tests

### Description

- Added an analytic test for RPCholesky on an independently derived example. 
- In the process, fixed a bug in the algorithm. 
- Also fixed a bug with the RPCholesky state (it now updates with the run of the algorithm) and the relating tests.
- Added a documentation page with the derivation and example code.

### How Has This Been Tested?

 Test A: `pytest tests/unit/test_solvers.py` passes. 
 Test B: `documentation/make html` builds the docs.

Not intentionally. 

- [x ] I have made sure that my PR is not a duplicate.
- [x ] My code follows the style guidelines of this project.
- [x ] I have commented my code, particularly in hard-to-understand areas.
- [x ] I have performed a self-review of my code.
- [x ] I have made corresponding changes to the documentation.
- [x ] My changes generate no new warnings.
- [x ] I have added tests that prove my fix is effective or that my feature works.
- [x ] New and existing unit tests pass locally with my changes.
- [x ] Any dependent changes have been merged and published in downstream modules.


gw265981:
### Note regarding unique/not-unique tests

 I believe RPCholesky will always choose unique points whether or not it is forced. In fact, forcing a choice of the same point seems to lead to a division by 0 in one of the steps, hence breaking the algorithm.

documentation/source/examples/analytical_rpcholesky.rst
```diff
@@ -0,0 +1,196 @@
...
+5. Update the residual diagonal:
+
+.. math::
+    d = d - |F[:,0]|^2 = \begin{pmatrix}
+    0.18126933 \\
+    0.39346947 \\
+    0
+    \end{pmatrix}
+
+For the second iteration (i=1):
```
qh681248, line 75: I am finding it hard to follow, could you add more explanation? I am not sure if I understand it correctly: We start with gram-matrix, we make choice and we update the column of the matrix, then we update the diagonal and that is our new gram matrix?  If that is the case, could you add what the updated gram matrix looks like after the first iteration here? And then we look at the diagonal again?

tests/unit/test_solvers.py
```diff
@@ -1048,15 +1050,238 @@ def solver_factory(self) -> Union[type[Solver], jtu.Partial]:
...
+    @pytest.mark.parametrize("use_cached_state", (False,))
+    def test_reduce(
+        self,
+        jit_variant: Callable[[Callable], Callable],
+        reduce_problem: _ReduceProblem,
+        use_cached_state: bool,
+    ) -> None:
+        """
+        Check `reduce` raises no errors.
+
```
qh681248, line 1063: Add
```
:param jit_variant:
:param reduce_problem:
:param use_cached_state: 
```
Check contributors guideline for the docstring conventions

tests/unit/test_solvers.py
```diff
@@ -1048,15 +1050,238 @@ def solver_factory(self) -> Union[type[Solver], jtu.Partial]:
...
+    @override
+    @pytest.mark.parametrize("use_cached_state", (False,))
+    def test_reduce(
+        self,
+        jit_variant: Callable[[Callable], Callable],
+        reduce_problem: _ReduceProblem,
+        use_cached_state: bool,
+    ) -> None:
+        """
+        Check `reduce` raises no errors.
```
qh681248, line 1062: If this refers to the `reduce` method of a class, it should be referenced here, for eg: `coreax.solvers.RPCholesky.reduce`

documentation/source/examples/analytical_rpcholesky.rst
gw265981: I hope the additional explanation of the algorithm in the beginning helps - let me know if it's still unclear

tests/unit/test_solvers.py
gw265981: I did not do this for now, happy to discuss the general strategy in the future

qh681248 (approved):
Great work!

qh681248 (approved):

PR #819, opened 2024-10-23
Release/0.3.0

qh681248:
### PR Type

- Release

### Description

This PR introduces version 0.3.0 of `Coreax`.
### How Has This Been Tested?

 Test A: Ran all tests on windows.
 Test B: Ran all the tests on the GPU.

### Does this PR introduce a breaking change?

No

### Screenshots

(Write your answer here.)

### Checklist before requesting a review
- [x] I have made sure that my PR is not a duplicate.
- [x] My code follows the style guidelines of this project.
- [x] I have commented my code, particularly in hard-to-understand areas.
- [x] I have performed a self-review of my code.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass locally with my changes.
- [x] Any dependent changes have been merged and published in downstream modules.

PR #836, opened 2024-10-31
fix: rpcholesky reverts to random selection after convergence

### LINKED ISSUE #835: RPCholesky stops adding points to the coreset
### What's the problem?

The RPCholesky (implemented in `coreax.solvers.RPCholesky.reduce()`) algorithm iteratively adds data points to the coreset by sampling according to the residual diagonal. This is done by iteratively building an approximation of the original Gram matrix and updating the residual diagonal to reflect which points have not been "covered". 

It is possible that this process converges before `coreset_size` has been reached. In this case, the residual diagonal becomes all 0 and so no new points are sampled (the rest of the points in the coreset are by default the 0-indexed one). 

I will add a small amount of noise to the residual diagonal (to the points that have not been selected before) to prevent this. Essentially, this emulates choosing random points after the RPCholesky procedure has converged. 

Run the `example.herding_basic.py` script with a relatively large coreset size (e.g., N=10000, m=1000) and observe the output of RPCholesky.

### Python version

3.9, 3.10, 3.11, 3.12

### Package version

0.2.1

### Operating system

Windows 11

### Other packages

_No response_

### Relevant log output

_No response_

gw265981:
### PR Type

- Bugfix

### Description

Make sure RPCholesky outputs a full coreset of `coreset_size` by falling back on random selection. Resolves #835.

### How Has This Been Tested?

- Tests pass
- No repeating points in the coreset, even if `coreset_size == num_data_points`. 
- The initial points before convergence coincide before and after the fix.

No.

### DIFF
--- a/coreax/solvers/coresubset.py
+++ b/coreax/solvers/coresubset.py
@@ -435,19 +435,28 @@ class RPCholesky(CoresubsetSolver[Data, RPCholeskyState], ExplicitSizeSolver):
         data. The pivot sampling and diagonal updating steps are repeated until
         :math:`M` points have been selected.
 
+        .. note::
+            The RPCholesky algorithm sometimes converges before reaching
+            ``self.coreset_size``. In this case, the other coreset points are chosen
+            uniformly at random, avoiding repetition.
+
+
         :param dataset: The dataset to reduce to a coresubset
         :param solver_state: Solution state information, primarily used to cache
             expensive intermediate solution step values.
         :return: a refined coresubset and relevant intermediate solver state information
         """
         x = dataset.data
+        num_data_points = len(x)
         if solver_state is None:
             gramian_diagonal = jax.vmap(self.kernel.compute_elementwise)(x, x)
         else:
             gramian_diagonal = solver_state.gramian_diagonal
-        initial_coresubset = _initial_coresubset(0, self.coreset_size, dataset)
+        # Initialise with an index outside of the possible coreset indices
+        initial_coresubset = _initial_coresubset(
+            num_data_points + 1, self.coreset_size, dataset
+        )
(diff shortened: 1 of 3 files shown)


coreax/solvers/coresubset.py
```diff
@@ -435,29 +435,53 @@ def reduce(
...
-        num_data_points = len(x)
 
         def _greedy_body(
             i: int, val: tuple[Array, Array, Array]
         ) -> tuple[Array, Array, Array]:
             """RPCholesky iteration - Algorithm 1 of :cite:`chen2023randomly`."""
             residual_diagonal, approximation_matrix, coreset_indices = val
             key = jr.fold_in(self.random_key, i)
+
+            # Fall back on uniform probability if residual_diagonal is invalid
```
gw265981, line 468: The problem with adding noise to simulate random choice is that the residual diagonal becomes `nan` anyway, so this stops working. Instead, we check if the diagonal is all 0 or has nans and then force random choice between the unpicked points.

gw265981:
Added a test that checks there are no duplicates in the coreset even if the full dataset is selected.

rg936672 (approved):
Looks good. The 20% performance hit from this isn't great, but correctness is more important than speed.

gw265981:
Yes, I'm definitely happy to work on optimisations if that ever becomes an issue, but I think for now it's more important to get RPCholesky to give a sensible output so it can be compared to other algorithms.

rg936672 (changes requested):
wait actually important change that we need to get in the habit of - add this to CHANGELOG! (You may want to wait until #819 is merged first though, to avoid further merge conflicts.)

rg936672 (approved):

Issue #810, opened 2024-10-18
Benchmarking: Image and Video data


qh681248:
### What's the new feature?

Use coreset algorithms to keep a subset of pixels (for image data) and subset of frames (for video data) and visually see how well they represent the original data

### What value does this add?

It adds another test for benchmarking and a rule of thumb for which algorithm to use for which task

bk958178:
Awaiting approval of #509

PR #838, opened 2024-11-01
Feature/image benchmarking

qh681248:
### PR Type

- Feature

### Description

Adds script to generate results from David and Pounce examples using several coreset algorithms.


qh681248:
Split into smaller PRs #810 and #892

Issue #847, opened 2024-11-06
Use headless version of opencv-python


tp832944:
### What's the issue?

When running from the command line (in a headless state), the package `opencv-python` needs to be replaced with `opencv-python-headless`. In fact, we should always do this because we do not require GUI features of opencv.

Add a comment on opencv versions to README in case the user wants to install a different, incompatible version - see https://pypi.org/project/opencv-python-headless/.

Issue #846, opened 2024-11-06
Add instructions to install CUDA extras in README


rg936672:
### What's the issue?

Currently, if you install Coreax on a Linux machine with an NVIDIA GPU, you get the following warning:

```
An NVIDIA GPU may be present on this machine, but a CUDA-enabled jaxlib is not installed. Falling back to cpu.
```

The solution to this is to install `jax` with the relevant CUDA extras, but how to do this is not immediately obvious from the warning (though a quick Google search does turn up some stackoverflow answers explaining it). It'd be good to have this issue specifically mentioned in the README so users don't have to separately search for it.

PR #848, opened 2024-11-06
Switch to headless opencv

tp832944:
Rework installation instructions in README.

closes: #846
closes: #847

### PR Type

- Bugfix
- Build related changes

### Description

Use package `opencv-python-headless`.

Rework installation instructions in README.

### How Has This Been Tested?

Tested on headless GPU machine.

Yes. Anyone using `opencv-python` will need to adjust their dependencies.


README.md
```diff
@@ -61,41 +61,48 @@ Run `examples/pounce.py` to replicate.
...
 
-Should the installation fail, try again using stable pinned package versions. Note that
-these versions may be rather outdated, although we endeavour to avoid versions with
-known vulnerabilities. To install Coreax:
+Should the installation of Coreax fail, try again using stable pinned package versions
+provided in `uv.lock`. First,
+[install UV](https://docs.astral.sh/uv/getting-started/installation/). Then, run
 ```shell
-$ python3 -m pip install --no-dependencies -r requirements.txt
+uv sync
```
rg936672, line 94: In order to use the pinned versions, an end user would first have to clone the repository (to get the `uv.lock`) and _then_ sync. Even then, that's not perfect as that installs it as a project, rather than as a dependency of the _user's_ project, which is presumably what they'd want. Instead, we should probably recommend the use of `uv export` to generate a requirements.txt file, which can then be installed using the user's choice of Python package manager - not 100% sure on what the cleanest way of doing that for a separate project is, though.

rg936672: (However, this particular change could also be considered out of scope for this PR - it might be worth splitting out some of the README update into a separate PR?)

rg936672 (approved):

PR #862, opened 2024-11-13
fix: mmd metric returning nan

gw265981:
`MMD^2` in the computation has been truncated to ensure it is non-negative.

### PR Type

- Bugfix

### Description

`MMD.compute` has been changed so that it returns 0 if precision error causes it to be negative.

### How Has This Been Tested?

The previous known case returns 0 instead of nan. 

No.

### DIFF
--- a/.cspell/library_terms.txt
+++ b/.cspell/library_terms.txt
@@ -74,6 +74,8 @@ meshgrid
 mimread
 mimsave
 minval
+MMD
+mmd
 modindex
 myst
 nabla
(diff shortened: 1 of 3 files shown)


.cspell/library_terms.txt
```diff
@@ -74,6 +74,8 @@ meshgrid
 mimread
 mimsave
 minval
+MMD
```
pc532627, line 77: I am 99% sure that this spell check is agnostic to case - if you remove `MMD` does pre-commit fail?

pc532627 (changes requested):
One minor comment

.cspell/library_terms.txt
gw265981: Removed it now, pretty sure it's fine.

pc532627 (approved):

Issue #688, opened 2024-07-05
Kernel Thinning


db091756:
### What's the new feature?

Add Kernel Thinning (https://arxiv.org/abs/2105.05842)

### What value does this add?

More features in Coreax.

tc85324:
I would be very interested to better understand the similarities between Kernel Thinning and "Map Reduce Kernel Herding". By applying MapReduce to Kernel Herding, I think we are doing a similar recursion to that described in Kernel Thinning. It might be that the only difference is in the "splitting" process. If that is the case it means we need only generalise MapReduce to allow arbitrary partitioning/splitting schemes.

PR #869, opened 2024-11-15
fix: correct refine behaviour on existing coreset

gw265981:
### PR Type

- Bugfix

### Description

Kernel Herding and Stein Thinning solvers implement a `refine` method which is meant to work with existing coresets. This functionality has been fixed so that starting with an existing coreset, each element is evaluated and potentially replaced according to the corresponding greedy rule.

### How Has This Been Tested?

The results for KH coincide with an independent numpy implementation that maximises MMD at every step.

No.


gw265981:
There are two improvements that can be made here:

1) Add analytic tests for refine for Kernel Herding and Stein Thinning.

2) Investigate whether the behaviour is correct for Stein Thinning. This is difficult to check since Stein Thinning has many moving parts. One thing to consider is whether the current iteration `i` should be replaced by current `coreset_size` in the entropic regularisation term. https://github.com/gchq/coreax/blob/98b61425d98e5321eb9ad4c4c6aa71b4295dd230/coreax/solvers/coresubset.py#L396

If this fix is immediately useful (e.g., for #688), I'm happy for this to be merged now and work on the improvements afterwards.

qh681248 (commented):
When passing a coreset, I can confirm that it produces the same output as minimising MMD at each step, I am happy with the changes, could you add a test that the current version would pass but the previous version would fail?

gw265981:
I will work on adding the test now.

Issue #744, opened 2024-08-22
Typing pass


db091756:
### What's the new feature?

Do a typing pass of the repo to ensure as many of the PyRight complaint are fixed as possible:

1. Look into `typing.overload`
2. Use `jax.typing` in as much as the codebase as possible to check array shapes

### What value does this add?

Consistency, better user experience

PR #878, opened 2024-11-24
feat: `jaxtyping` support for `coreax.data.Data`

### LINKED ISSUE #765: Look into typing `Data` and `SupervisedData` instances

tc85324:
Closes #765 -- if this PR is merged, it will enable increased usage of `jaxtyping` (desired in #744).

### PR Type
- Feature
- Tests

### Description
Enables `Data` and `SupervisedData` instances to be given a type hint that indicates the type and size of their respective `Data.data` attributes.

For example: `x: Int[Data, "n d"] = ...`, indicates `x` is an instance of `Data` whose `data` attribute is an integer array of shape `n d`.

### How Has This Been Tested?

New tests introduced with this change verify that (runtime) typing works as expected.

Introduces a lower bound on the required version of `jaxtyping`.


tests/unit/test_data.py
```diff
@@ -160,6 +173,53 @@ def test_normalize(self, data_type, weights):
...
+        for data in [int_data, float_data]:
+            good_shape_check(data)
+            with bad_shape_check_context:
+                bad_shape_check(data)
+            # We don't expect an error to be raised in the `float_data`` case, on a call
+            # to float_dtype_check(...)
+            if data is float_data:
+                float_check_context = nullcontext()
+            with float_check_context:
+                float_dtype_check(data)
```
rg936672, line 221: For this entire section, rather than defining functions to check with `@jaxtyped` and checking if they error, it'd be much simpler to just use `assert beartype.door.is_bearable(data, type_hint)` (or `assert not`, as required) to check whether or not a type hint matches.

tests/unit/test_data.py
```diff
@@ -147,6 +150,16 @@ def test_len(self, data_type):
...
 
+    def dtype(self, data_type):
+        """Test dtype property; required for jaxtyping annotations."""
+        _data = data_type()
+        assert _data.data.dtype == _data.dtype
+
+    def shape(self, data_type):
+        """Test shape property; required for jaxtyping annotations."""
+        _data = data_type()
+        assert _data.data.shape == _data.shape
```
rg936672, line 161: If you want these to be run as tests, you need to ensure their names start with `test_` else Pytest won't run them

tests/unit/test_data.py
```diff
@@ -99,8 +102,8 @@ def test_atleast_2d_consistent(arrays: tuple[Array]) -> None:
 @pytest.mark.parametrize(
     "data_type",
     [
-        partial(coreax.data.Data, DATA_ARRAY),
-        partial(coreax.data.SupervisedData, DATA_ARRAY, SUPERVISION),
+        jtu.Partial(coreax.data.Data, DATA_ARRAY),
+        jtu.Partial(coreax.data.SupervisedData, DATA_ARRAY, SUPERVISION),
```
rg936672, line 106: Happy for this to be out of scope, but it might be nice to wrap these in a `pytest.param` each to give them a friendly name for the logs.

tests/unit/test_data.py
```diff
@@ -160,6 +173,53 @@ def test_normalize(self, data_type, weights):
...
+        float_data_type = eqx.tree_at(
+            lambda x: x.args,
+            data_type,
+            replace=jtu.tree_map(lambda y: jnp.astype(y, jnp.float32), data_type.args),
+        )
+        int_data_type = eqx.tree_at(
+            lambda x: x.args,
+            data_type,
+            replace=jtu.tree_map(lambda y: jnp.astype(y, jnp.int32), data_type.args),
+        )
```
rg936672, line 194: We could instead parametrise the dtype for this test here, to avoid code duplication

tests/unit/test_data.py
```diff
@@ -160,6 +173,53 @@ def test_normalize(self, data_type, weights):
...
         )
 
+    def test_jaxtyping_compatibility(self, data_type):
+        """
+        Test `Data` compatibility with jaxtyping annotations.
+
+        Checks the following cases:
+            - Correct shape and data type,
+            - Correct shape and incorrect data type,
+            - Incorrect shape and correct data type.
```
rg936672, line 183: Could maybe also parametrise the shape we check as well, but that might be less clear than doing two separate checks

pyproject.toml
```diff
@@ -30,7 +30,7 @@ dependencies = [
     "flax",
     "jax",
     "jaxopt",
-    "jaxtyping",
+    "jaxtyping>0.2.31",
```
rg936672, line 33: Is this a breaking change? Or did we already implicitly depend on the version being above 0.2.31 and just didn't have a lower bound listed here?

rg936672 (changes requested):
This is a good change! A couple of issues in the tests, and some suggestions on parametrising them.

rg936672:
Also, to fix the merge conflicts, merge in the latest `main`, resolving the merge conflicts however you like (easiest is just to blindly use one of the two files rather than trying to actually resolve them), and then run `uv sync --upgrade` to update the uv.lock.

tests/unit/test_data.py
tc85324: Don't know how I missed this, thanks for spotting!

pyproject.toml
tc85324: This PR explicitly requries the version to be above 0.2.31 (the functionality in this PR breaks on lower versions). I'm not sure what the implicit lower bound was before.

tests/unit/test_data.py
tc85324: I like this idea! Not sure why I made this so complicated first time round.

tests/unit/test_data.py
tc85324: I think it is clearer to do these checks separately.

pyproject.toml
rg936672: In which case, it's probably safer to log this as a breaking change in the CHANGELOG.

PR #879, opened 2024-11-25
fix: correct refine behaviour on existing coreset

gw265981:
Reupload of #869 due to a typo in the branch name. See original PR for discussion.

### PR Type

- Bugfix

### Description

Kernel Herding and Stein Thinning solvers implement a `refine` method which is meant to work with existing coresets. This functionality has been fixed so that starting with an existing coreset, each element is evaluated and potentially replaced according to the corresponding greedy rule.

### How Has This Been Tested?

The results for KH coincide with an independent numpy implementation that maximises MMD at every step.

No.


gw265981:
There is an edge case when setting `unique=True`. This implementation of refine selects the best candidate point that is different from the points that come _before_ in the coreset. So it may select a point that is already in the original coreset if that point is later in the order. Of course, if we are refining the whole coreset, that point will later be replaced to be unique, so this always yields a unique coreset as required. 

However, if we support partial refinement in the future, this might be undesirable. E.g., if we are only replacing the first 5 out of 10 points, one of the new points may be the same as a point in the latter 5 that are not replaced. I will change the refine method accordingly so that each selected point is unique with respect to the whole existing coreset.

qh681248 (dismissed):
Thank you for updating the refine method and adding the unit test. I would say that "The test case has been verified independently to minimise MMD at every iteration." tells us what we should do in case a numpy update breaks all random seeds. I am happy with the changes. Happy to merge after the merge conflict is resolved.

qh681248 (approved):
Thanks

qh681248 (approved):

Issue #807, opened 2024-10-14
Add lower bounds on dependencies


rg936672:
### What's the issue?

Currently, most of Coreax's dependencies don't have lower bounds specified, so aren't accurate - we clearly don't support very early versions of most of our dependencies. Add minimum versions for each, ideally as loose as possible. This can be tested with uv's `--resolution lowest` strategy. - see #918.

Since automated testing of this is a separate ticket, this ticket can be resolved by just adding naive "latest major version" bounds on the dependencies. (Just doing this puts us in a much better position than the majority of the Python ecosystem!)

pc532627:
Blocked by #509

PR #898, opened 2024-12-16
Bugfix/score matching equinox upgrade

gw265981:
### PR Type

- Bugfix
- Build related changes

### Description

- Small change to the ScoreMatching class to make it work with equinox>0.11.7.
- Updated version constraint for equinox as it no longer needs to bounded from above.

### How Has This Been Tested?

Unit tests now pass.

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Moved coverage and performance data from GitHub gist to coreax-metadata repo.
   (https://github.com/gchq/coreax/pull/887)
+- Equinox dependency version is changed from `<0.11.8` to `>=0.11.9`. (https://github.com/gchq/coreax/pull/898)
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,7 +41,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Moved coverage and performance data from GitHub gist to coreax-metadata repo.
   (https://github.com/gchq/coreax/pull/887)
-- Equinox dependency version is changed from `<0.11.8` to `>=0.11.9`. (https://github.com/gchq/coreax/pull/898)
+- **[BREAKING CHANGE]** Equinox dependency version is changed from `<0.11.8` to `>=0.
+11.5`. (https://github.com/gchq/coreax/pull/898)
 
 ### Removed
 
--- a/coreax/score_matching.py
+++ b/coreax/score_matching.py
@@ -62,12 +62,15 @@ _RandomGenerator = Callable[[KeyArrayLike, Sequence[int], DTypeLike], Array]
 
 class ScoreMatching(eqx.Module):
     """
-    Base class for score matching algorithms.
+    Abstract base class for score matching algorithms.
 
     The score function of some data is the derivative of the log-PDF. Score matching
     aims to determine a model by 'matching' the score function of the model to that
     of the data. Exactly how the score function is modelled is specific to each
     child class of this base class.
+
+    This class should only contain abstract methods. Subclasses must implement all
+    abstract methods to create concrete score matching algorithms.
     """
 
     @abstractmethod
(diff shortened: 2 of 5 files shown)


coreax/score_matching.py
```diff
@@ -60,7 +60,7 @@
 _RandomGenerator = Callable[[KeyArrayLike, Sequence[int], DTypeLike], Array]
 
 
-class ScoreMatching(ABC, eqx.Module):
```
rg936672, line 63: I don't get any test errors even without this change? (With the versions specified in the `uv.lock` on this branch.)

CHANGELOG.md
```diff
@@ -41,6 +41,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Moved coverage and performance data from GitHub gist to coreax-metadata repo.
   (https://github.com/gchq/coreax/pull/887)
+- Equinox dependency version is changed from `<0.11.8` to `>=0.11.9`. (https://github.com/gchq/coreax/pull/898)
```
rg936672, line 44: This is technically a breaking change, and we should highlight it as such!

coreax/score_matching.py
gw265981: Interesting, I think it fails if ABC is the first argument (like it was before) but not if it's the second one. I'm happy to add ABC as a second argument if there is a good reason for this?

CHANGELOG.md
gw265981: Is there a special format for this? Also, the contributor guidelines state that we need a deprecation period for breaking changes, but it seems like overkill in this instance.

rg936672: I don't think there's a specific format for it; [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) doesn't specify one. I'd just add `**[BREAKING CHANGE]**` to the start of the line to make it obvious.

I agree that a deprecation period seems overkill here, especially since there's not really a way to have a transition period.

However, I'm not sure we actually need to limit the Equinox version so tightly in the first place - the tests seem to run fine with Equinox versions older than `0.11.9` (e.g. `0.11.7`)? We should definitely have _some_ lower bound on the Equinox version though - e.g. I'd imagine Coreax won't work with `equinox==0.0.1`!

coreax/score_matching.py
gw265981: I was just following the example from equinox: https://docs.kidger.site/equinox/pattern/, who don't include ABC in the inheritance. I have tested that `@abstractmethod` is enough to enforce method implementation, I'm not sure if there is any other functionality we want. 

If it's just for the purpose of being explicit, I'm happy to include the explicit inheritance, although the decorator is fairly explicit. 

Anyway, this is a fairly insignificant change, so happy to go with your final call, but it's been a good discussion in case something like this crops up again.

CHANGELOG.md
gw265981: From my very rigorous testing, the tests fail for 0.11.0 but pass for 0.11.5, so the latter should be relatively safe. A potential issue is that the dev environment will often resolve to the latest version so we won't really know if something breaks for the older versions.

coreax/score_matching.py
rg936672: Having looked a little closer into the code, `eqx.Module` is itself an abstract class (its metaclass is a subclass of `ABCMeta`) so we shouldn't need to inherit from ABC or explicitly set the metaclass ourself.

I think we'd want to do at least one of: 
- renaming it to `AbstractScoreMatching` (this one is what the Equinox docs do, but renaming is a fairly broad change which might be a hassle wrt deprecation periods etc)
- Making it very obvious in the docstring (in the first line ideally) that it's an _abstract_ base class

to make it immediately obvious that it's an abstract class.

CHANGELOG.md
rg936672: In which case, the latter is a good lower bound. Since tests fail for earlier versions, I think bumping the minimum required version isn't a breaking change _per se_, since it's just making a pre-existing requirement explicit, but it's still worth highlighting in the changelog.

Checking our lower bounds are correct is something that we can do by setting uv's [`--resolution-strategy`](https://docs.astral.sh/uv/concepts/resolution/#resolution-strategy) to `lowest` (or possibly `lowest-direct` in a pinch) - I'm planning on adding this in a CI job at some point, either as part of resolving #807 or as a subsequent PR.

rg936672 (approved):
Approved, assuming the CI passses.

PR #899, opened 2024-12-17
Pylint pre-commit should not run in parallel

rg936672:
### PR Type

- Bugfix
- CI related changes

### Description

The Pylint docs recommend against using Pylint as a pre-commit hook in the first place, but that if you do need to set `require_serial` to `true`. Otherwise, it might give inconsistent results on some checkers (such as `duplicate-code`) since the files that contain duplicate code might be analysed by two separate Pylint processes.

See: https://pylint.pycqa.org/en/latest/user_guide/installation/pre-commit-integration.html

### How Has This Been Tested?

Pre-commit still runs as expected; testing the effects directly is not easy but this should avoid inconsistent results on certain checkers.

No

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -121,7 +121,7 @@ repos:
         name: pylint
         entry: pylint
         language: system
-        require_serial: true  # ensure only on Pylint process is used
+        require_serial: true  # ensure only one Pylint process is used
         types:
           - python
         args:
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -121,6 +121,7 @@ repos:
         name: pylint
         entry: pylint
         language: system
+        require_serial: true  # ensure only on Pylint process is used
```
qh681248, line 124: Small typo `on` should be `one`, but happy with the changes

qh681248 (changes requested):

qh681248 (approved):

PR #909, opened 2024-12-24
Typing and build fixes to benchmarking

### LINKED ISSUE #744: Typing pass

ghost:
### PR Type

- Bugfix
- Build related changes

### Description

Add extra dependencies to `benchmark` in pyproject.toml to match what scripts require.

Add `example` dependency group.

Additional typing and refactoring in benchmark scripts. No idea how the original worked! This should be clearer and pass pyright.

Works towards but does not close #744.

### How Has This Been Tested?

pylint passes. Unchecked on pyright.

No.


benchmark/mnist_benchmark.py
```diff
@@ -592,23 +594,21 @@ def save_results(results: dict) -> None:
...
+    :param solver: An instance of a solver, such as `MapReduce` or `RandomSample`.
     :return: The name of the solver class.
     """
-    solver_name = (
-        _solver.base_solver.__class__.__name__
-        if _solver.__class__.__name__ == "MapReduce"
-        else _solver.__class__.__name__
-    )
-    return solver_name
+    if (rtn_type := signature(solver).return_annotation) is MapReduce:
```
qh681248, line 608: `signature(an instance of RandomSample)` gives the following error
`TypeError: RandomSample(coreset_size=25, random_key=u32[2], weighted=False, unique=True) is not a callable object`, this could be because `solver` is a `Solver` and not a `Callable`

qh681248: The current fix is returning _ActualModuleMeta` for `RandomSample` and `RPCholesky` but it is returning the correct name for the solvers that use `MapReduce`

benchmark/mnist_benchmark.py
```diff
@@ -592,23 +594,21 @@ def save_results(results: dict) -> None:
...
+        # Evaluate solver function to get MapReduce instance
+        return solver(1).base_solver.__class__.__name__
+    return rtn_type.__class__.__name__
```
qh681248, line 611: Should the return be `solver(1).__class__.__name__`?

pyproject.toml
qh681248, line 1: `opencv-python-headless` is needed to run `david_benchmark.py` and `david_map_reduce_weighted.py`, do you think it is worth adding in benchmarking dependencies/main dependencies?

tp832944: More widespread updates to dependencies implemented.

benchmark/mnist_benchmark.py
tp832944: Fixed.

benchmark/mnist_benchmark.py
tp832944: Rewritten to be more robust.

qh681248 (approved):
Happy with all the changes

PR #964, opened 2025-02-10
Feature/backup compress++

qh681248:
#934

This pr replaces #939 
### PR Type

- Feature

### Description

Implementation of Compress plus plus algorithm from   https://arxiv.org/pdf/2111.07941.pdf using kernel thinning

### How Has This Been Tested?

New unit tests being added

No

### DIFF
--- a/coreax/solvers/coresubset.py
+++ b/coreax/solvers/coresubset.py
@@ -1301,3 +1301,144 @@ class KernelThinning(CoresubsetSolver[_Data, None], ExplicitSizeSolver):
             coreset_size=self.coreset_size, kernel=self.kernel
         ).refine(candidate_coreset)
         return refined_coreset, None
+
+
+class CompressPlusPlus(CoresubsetSolver[_Data, None], ExplicitSizeSolver):
+    r"""
+    Compress++ - A hierarchical coreset construction solver.
+
+    `CompressPlusPlus` is an efficient method for building coresets without
+    compromising performance significantly. It operates in two steps: recursively
+    halving and thinning.
+
+    In the recursive halving step, the dataset is partitioned into four subsets.
+    Each subset is further divided into four subsets, repeated a predefined number
+    of times. After this, each subset is halved using `~coreax.solvers.KernelThinning`.
+    The halved subsets are concatenated bottom-up to form a coreset.
+
+    Finally, the resulting coreset is thinned again using
+    `~coreax.solvers.KernelThinning` to obtain a coreset of the desired size. This
+    implementation is an adaptation of the Compress++ algorithm in
+    :cite:`shetty2022compress` to make it an explicit sized solver.
+
+    :param g: The oversampling factor.
+    :param kernel: A `~coreax.kernels.ScalarValuedKernel` for kernel thinning.
+    :param random_key: A random number generator key for the kernel thinning solver,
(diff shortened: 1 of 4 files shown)


coreax/solvers/coresubset.py
```diff
@@ -1301,3 +1300,154 @@ def kt_refine(
...
+        if not self.coreset_size <= 2**self.g * math.sqrt(effective_data_size):
+            raise ValueError(
+                "Coreset size and g are not compatible with the dataset size."
+            )
+        # Clip the dataset to the nearest power of 4
+        clipped_indices = jax.random.choice(
+            self.random_key, n, shape=(effective_data_size,), replace=False
+        )
+
+        def _compress_half(indices: jax.Array) -> jax.Array:
```
gw265981, line 1380: Maybe add a comment here and in `_compress_thin` that, in general, any halving and thinning functions can be used, and Kernel Thinning is used as it's a good coreset method.

coreax/solvers/coresubset.py
```diff
@@ -1301,3 +1300,154 @@ def kt_refine(
...
+            data_to_half = dataset[indices]
+            thinning_solver = KernelThinning(
+                coreset_size=m // 2,
+                delta=self.delta,
+                kernel=self.kernel,
+                sqrt_kernel=self.sqrt_kernel,
+                random_key=self.random_key,
+            )
+            halved_coreset, _ = thinning_solver.reduce(data_to_half)
+            return indices[halved_coreset.indices.data.flatten()]
```
gw265981, line 1398: I think `halved_coreset.unweighted_indices` still works. (Not necessary to change but might be convenient for the future)

coreax/solvers/coresubset.py
gw265981, line 1380: I believe `jax.Array` and `Array` from jaxtyping are equivalent here, but the latter is used in the rest of the code, so replace it for consistency (here and in the other methods).

coreax/solvers/coresubset.py
```diff
@@ -1301,3 +1300,154 @@ def kt_refine(
...
+        :param dataset: The original dataset to be reduced.
+        :param solver_state: The state of the solver.
+        :return: A tuple containing the final coreset and the solver state (None).
+        """
+        n = len(dataset)
+        if self.coreset_size > len(dataset):
+            raise ValueError(MSG)
+
+        # Check that depth and coreset_size are compatible
+        nearest_power_of_4 = math.floor(math.log(n, 4))
```
gw265981, line 1363: Replace `math` references with `jnp` throughout the file (I notice I've missed them in KernelThinning too so feel free to do those as well). This shouldn't change anything but better for consistency.

qh681248: Left it as it is because I get tracer error of the form:

```"""Copy an element of an array to a standard Python scalar and return it."""
>     arr = core.concrete_or_error(np.asarray, self, context="This occurred in the item() method of jax.Array")
E     jax.errors.ConcretizationTypeError: Abstract tracer value encountered where concrete value is expected: traced array with shape int32[]
E     This occurred in the item() method of jax.Array
 ```
 
 But happy to investigate it further if necessary

gw265981 (dismissed):
Just fix the changelog to stop the CI complaining :)

gw265981 (approved):

Issue #985, opened 2025-02-21
Add better alt texts for all the images in the benchmarking rst file


qh681248:
### What's the new feature?

Alternative text should be more descriptive. Add better alt texts for all the images in the benchmarking rst file.

### What value does this add?

Improves accessibility of documentation

gw265981:
Blocked by #977.

PR #1000, opened 2025-03-18
Bugfix/investigate stein thinning

gw265981:
### PR Type

- Bugfix
- Feature

### Description

Improving the performance of Stein Thinning in benchmarks, primarily by using regularisation and better probability density estimation.

### How Has This Been Tested?


gw265981:
List of all changes:

- [ ] Replace the score function for Stein Thinning in `blobs_benchmark.py` and `benchmark_util.py` (and so for other benchmarks) with `jax.scipy.stats.gaussian_kde()`. Also set regularisation to True.
- [ ] Add a parameter `kde_bw_method` to the SteinThinning solver which is passed to `jax.scipy.stats.gaussian_kde()` when estimating probability density for regularisation. 
- [ ] Small updates to benchmarks, including making sure the saving paths are correct, fixing random state of UMAP for reproducibility, reducing UMAP dimension for the pounce benchmark.
- [ ] Rerun all the benchmarks and update the benchmarking.rst page.
- [ ] Add test to `test_benchmark.py` to run the initialised SteinThinning solver.

gw265981:
Also addressed #992, #985. For the alt text, I left it unchanged for the blobs benchmark charts since raw data is provided.

benchmark/david_benchmark.py
```diff
@@ -28,8 +28,14 @@
 Each coreset algorithm is timed to measure and report the time taken for each step.
 """
 
-import math
 import os
+import sys
+
+sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), "..")))
```
rg936672, line 34: Does it make sense to just move `downsample_opencv` into `coreax.benchmark_util` rather than doing sys.path hackery here?

benchmark/david_benchmark.py
```diff
@@ -44,15 +50,19 @@
...
 
-# pylint: disable=too-many-locals, too-many-statements
+# pylint: disable=too-many-locals
 # ruff: noqa: PLR0914, PLR0915
 # (disable line too long and too many statements ruff)
 def benchmark_coreset_algorithms(
     in_path: Path = Path("../examples/data/david_orig.png"),
-    out_path: Optional[Path] = Path("david_benchmark_results.png"),
+    out_path: Optional[Path] = Path(
+        "../examples/benchmarking_images/david_benchmark_results.png"
```
rg936672, line 64: Should clarify in the docstring that this is relative to `__file__` if not given as an absolute path, as this would probably be unexpected.

benchmark/pounce_benchmark.py
```diff
@@ -35,14 +35,17 @@
 from coreax.data import Data
 
 
-def plot_selected_frames(umap_data, selected_indices, action_frames, solver_name):
+def plot_selected_frames(
+    umap_data, selected_indices, action_frames, solver_name, out_dir
```
rg936672, line 39: Should have type hints for arguments and return value here!

coreax/benchmark_util.py
```diff
@@ -195,14 +196,30 @@ def _get_stein_solver(_size: int) -> Union[SteinThinning, MapReduce]:
...
+            """
+
+            def logpdf_single(x: Shaped[Array, " d"]) -> Shaped[Array, ""]:
+                return kde.logpdf(x.reshape(1, -1))[0]
+
+            return jax.grad(logpdf_single)(x)
+
+        stein_kernel = SteinKernel(
+            base_kernel=kernel,
+            score_function=score_function,
```
rg936672, line 219: This code seems near-identical to that in `blobs_benchmark` - does it make sense to factor it out into a separate function? (Happy for the answer to be "no" here!)

rg936672 (changes requested):
Looks good, just a few minor suggestions. Remember to link the issues that this PR closes, either under "Development" at the top right, or by using "Closes #xyz" in the issue description.

benchmark/david_benchmark.py
gw265981: This sounds good to me; I assume duplicating code is not a big issue between different "packages". However, I was holding off to have a separate ticket for a rework of benchmarking which would include these kinds of changes. 

This change in particular was only included since it was already necessary to run the code but wasn't specified before, so I think it's better to make it explicit.

benchmark/david_benchmark.py
gw265981: Yep, will do. Again, these kinds of things need to be made more robust in general, but the priority for this ticket was to finish the Stein Thinning changes quickly.

benchmark/pounce_benchmark.py
gw265981: Again, completely agree, but would rather add this to the benchmark rework list for a new ticket.

coreax/benchmark_util.py
gw265981: Ideally, we would replace the solver setup code in blobs_benchmark.py with initialise_solvers from here.

benchmark/blobs_benchmark.py
```diff
@@ -84,17 +87,32 @@ def setup_stein_kernel(
...
+    kde = jsp.stats.gaussian_kde(dataset.data[idx].T)
+
+    # Define the score function as the gradient of log density given by the KDE
+    def score_function(
+        x: Union[Shaped[Array, " n d"], Shaped[Array, ""], float, int],
+    ) -> Union[Shaped[Array, " n d"], Shaped[Array, " 1 1"]]:
+        """
+        Compute the score function (gradient of log density) for a single point.
+
+        :param x: Input point represented as array
```
pc532627, line 104: Full stop at the end of this line and the following one

rg936672 (dismissed):
I'm happy my suggestions have all been addressed, though hold off on merging until both (a) @pc532627 approves as well and (b) you've created a ticket for any suggestions that you've deemed out of scope for this PR. (Could be as simple as a "Clean up benchmarking code" with a link to the review comments in question.)

pc532627 (approved):

documentation/source/benchmark.rst
```diff
@@ -58,23 +61,31 @@ The benchmarking test showed that the accuracy remained similar regardless of
...
-
 .. image:: ../../examples/benchmarking_images/mnist_benchmark_accuracy.png
- :alt: Benchmark Results for MNIST Coreset Algorithms
+ :alt: A bar chart showing the accuracy of nine different algorithms
+    (the 6 mentioned above and 3 variants of iterative probabilistic Kernel Herding) across
+    six coreset sizes (25, 50, 100, 500, 1000, and 5000). The chart displays increasing
+    accuracy for all algorithms as coreset size increases, with performance ranging from
+    about 0.45-0.55 at size 25 to about 0.9 at size 5000. There is
+    little difference in performance between the algorithms at the 1000-5000 size. No
+    algorithm can be said to outperform the random sample consistently and Stain Thinning
```
pc532627, line 76: @gw265981 I think there is a typo here `Stain` -> `Stein`

pc532627 (changes requested):
One minor typo

rg936672:
TODO failure is a bug in the latest Ruff - see https://github.com/astral-sh/ruff/issues/16874

pc532627 (approved):

PR #1022, opened 2025-04-09
ci: remove `pyright.yml`.

tm91236:
### PR Type
- CI related changes

### Description
Replace the functionality provided in `pyright.yml` with the `pre_commit_checks.yml` workflow by adding `pyright` to the `.pre-commit-config.yaml`.

This has the additional benefit of informing users of the CI expectations in pre-commit.

### How Has This Been Tested?
N/A

N/A

- [ ] I have ensured my code is easy to understand, including docstrings and comments where necessary.


gchqdev227 (commented):
I understand not wanting to maintain a custom action for `pyright`, but looking at [this thread](https://github.com/microsoft/pyright/issues/3612), it looks like there is an existing and well-supported GitHub Action recommended [here](https://github.com/jakebailey/pyright-action). Have we considered whether using that instead of a local `pre-commit` hook would be more appropriate?

tm91236:
Something I should have made clearer in the original PR description is that the primary reason to remove this workflow and move the pyright check into the pre-commit (which has its own CI workflow), is that it provides a developer with better visibility into the CI outcome at commit time. We wouldn't get this benefit by using the GitHub action.

gchqdev227 (approved):

PR #1088, opened 2025-11-21
fix: KernelThinning partitioning.

### LINKED ISSUE #1086: KernelThinning does not seem to partition correctly
### What's the problem?

In `KernelThinning.kt_half_recursive` the block 

```
if m == 0:
            return [
                Coresubset(Data(jnp.arange(len(current_coreset))), original_dataset)
            ]
```

means that each of the final partitions is returning a `Coresubset` that just consits of indices [0, 1, 2, ..., `coreset_size`] . Replacing it with 

```
if m == 0:
            return [current_coreset]
```

is the intended behaviour I believe.

Run `kt_half_recursive` for any dataset, and check the `.unweighted_indices` of each partition. They should all be identically [0, 1, ..., `coreeset_size`].

### Python version

3.12

### Package version

1.0.0

### Operating system

Ubuntu 24.04

### Other packages

_No response_

### Relevant log output

```shell

```

tm91236:
### PR Type
- Bugfix
- Tests

### Description
Closes #1086

### How Has This Been Tested?
Adds a test to verify that `kt_half_recursive` behaves as expected. Previously, `kt_half_recursive` would return a list of identical coresets, rather than a recursive split.

No.

### DIFF
--- a/tests/unit/test_solvers.py
+++ b/tests/unit/test_solvers.py
@@ -3370,6 +3370,10 @@ class TestKernelThinning(ExplicitSizeSolverTest):
         result = thinning_solver.kt_half_recursive(dataset, m, dataset)
         expected_result = [Coresubset(Data(jnp.arange(dataset_size)), dataset)]
         assert eqx.tree_equal(expected_result, result)
+        coresubset = Coresubset(Data(jnp.array([1, 4, 3])), dataset)
+        result = thinning_solver.kt_half_recursive(coresubset, m, dataset)
+        expected_result = [coresubset]
+        assert eqx.tree_equal(expected_result, result)
 
 
 class TestCompressPlusPlus(ExplicitSizeSolverTest):


coreax/solvers/coresubset.py
```diff
@@ -1401,10 +1401,10 @@ def kt_half_recursive(
         """
         # If m == 0, do not do anything just convert to original data to type Coresubset
         if m == 0:
-            return [
-                Coresubset(Data(jnp.arange(len(current_coreset))), original_dataset)
-            ]
-
+            if isinstance(current_coreset, Coresubset):
+                return [current_coreset]
```
gchqdev227: Do you have a test for this case, where you assert that `result[0] is current_coreset`?

gchqdev227: That would make sense, but even so the test would pass if we returned a new "copy" of `current_coreset`, which is presumably what this fix is trying to avoid? The specific behaviour here is that `current_coreset` is being returned as-is, which is much more efficient.

tm91236: No problem! I've extended the existing test to explicitly cover this case as it improves the test quality.

tests/unit/test_solvers.py
```diff
@@ -3370,6 +3370,10 @@ def test_kt_half_recursive_zero_depth(self, solver_factory: jtu.Partial):
         result = thinning_solver.kt_half_recursive(dataset, m, dataset)
         expected_result = [Coresubset(Data(jnp.arange(dataset_size)), dataset)]
         assert eqx.tree_equal(expected_result, result)
+        coresubset = Coresubset(Data(jnp.array([1, 4, 3])), dataset)
+        result = thinning_solver.kt_half_recursive(coresubset, m, dataset)
+        expected_result = [coresubset]
+        assert eqx.tree_equal(expected_result, result)
```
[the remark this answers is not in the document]
tm91236: This assertion is pretty much equivalent to `assert core_subset == expected_result[0]`. IMO concern over copying the `Coresubset`, or more precisely the JAX arrays that are leaves of the coresubset, can be deferred to the JIT compiler (which should avoid such behaviour when feasible). 

The additional test case does explicitly check that `kt_half_recursive` operates properly on a `Coresubset` input, but this was already implicit before.

PR #858, opened 2024-11-12
Automated build testing

rg936672:
### PR Type

- Build related changes
- CI related changes

### Description

Closes #854.

Adds a manually-triggered Action that builds the package, and then installs and tests the built artifact across all our supported Python/OS combinations.

Currently, this does _not_ create a GitHub release, or upload the built artifact to PyPi, or build the documentation for ReadTheDocs. Those would be good things to add to this action in future, though - the more automated the build and release process is, the better!

### How Has This Been Tested?

See https://github.com/rg936672/coreax/actions/runs/11798590544 for a successful run.

### Does this PR introduce a breaking change?

No.

### Checklist before requesting a review
- [x] I have made sure that my PR is not a duplicate.
- [x] My code follows the style guidelines of this project.
- [x] I have ensured my code is easy to understand, including docstrings and comments where necessary.
- [x] I have performed a self-review of my code.
- [x] I have made corresponding changes to the documentation.
- [x] My changes generate no new warnings.
- [x] New and existing unit tests pass locally with my changes.
- [x] Any dependent changes have been merged and published in downstream modules.
- [x] I have updated CHANGELOG.md, if appropriate.

PR #1150, opened 2026-03-12
ci: Enable trusted publishing release workflow.

tm91236:
Closes #858

### PR Type

- Build related changes
- CI related changes

### Description
Enables a semi-automated release process.

### How Has This Been Tested?
N/A

No

- [] I have made corresponding changes to the documentation.

### DIFF
--- a/.cspell/custom_misc.txt
+++ b/.cspell/custom_misc.txt
@@ -57,6 +57,7 @@ primaryclass
 PRNG
 prob
 propto
+pypa
 rcond
 recomb
 refs
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,12 +1,9 @@
 name: Release
 
 on:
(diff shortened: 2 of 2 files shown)


sylvesterkaczmarek (commented):
The tag ruleset protects the source test matrix, but this change removes a separate release-safety property from the old workflow: the built wheel was installed into a sparse checkout and tested across Python 3.10-3.14 on Linux, Windows and macOS. The replacement only imports the wheel/sdist on Python 3.14/Linux before publishing to PyPI.

That can miss packaging defects where source-tree tests pass but files/modules/package data are absent from the wheel, as well as supported-version packaging failures. Before the publish step, could we retain an artefact-level test job (at least across the supported Python versions, ideally the existing OS matrix) that installs only the built wheel and runs the tests against it? The trusted-publishing/provenance direction itself looks good.

PR #1235, opened 2026-09-07
test: improve metric assertions

### LINKED ISSUE #1140: Improve metric tests
### What's the new feature?

Improve the tests quality of `test_metrics.py`. Tests that copy the implementation should be replaced with tests that assert known properties of the solution.

### What value does this add?

Removes tests that artificially increase coverage and encourages higher quality test patterns.

_No response_

### Additional context

_No response_

sylvesterkaczmarek:
### PR Type

- Tests

### Description

Improves the quality of the metric tests identified in #1140.

- Replaces the random-data MMD test that recomputed the MMD formula with checks for symmetry and sample-order invariance, for both weighted and unweighted data.
- Replaces the random-data KSD test that recomputed KSD internals with sample-order invariance checks across unweighted, weighted, Laplace-corrected and regularised modes.
- Removes imports that were only needed by the duplicated test implementations.
- Records the test-quality change in the changelog.

Closes #1140.

### How Has This Been Tested?

- `pytest tests/unit/test_metrics.py`: 38/38 passed.
- `pytest tests/`: 1,754 passed, 19 skipped.
- `pre-commit run --all-files`: passed, including Ruff, Pyright, Pylint and cspell.
- `git diff --check`: passed.

No.


tests/unit/test_metrics.py
```diff
@@ -338,70 +325,55 @@ def _score_function(x: ArrayLike) -> Array:
...
+
+        actual = metric.compute(
+            permuted_reference,
+            permuted_comparison,
+            laplace_correct=laplace_correct,
+            regularise=regularise,
+        )
+        assert actual == pytest.approx(expected, abs=1e-6, rel=1e-3)
+        assert jnp.isfinite(actual)
+        assert actual >= 0
```
tm91236: Would we ever expect the assertions on lines 375 and 376 to fail if the assertion on line 374 passes?

tests/unit/test_metrics.py
```diff
@@ -193,47 +190,37 @@ def test_mmd_analytically_known_weighted(self) -> None:
...
+                comparison_data.data[::-1], jnp.asarray(comparison_data.weights)[::-1]
             )
         else:
-            raise ValueError("Invalid mode parameterization")
-        # Compute the MMD using the metric object
-        metric = MMD(kernel=kernel)
-        output = metric.compute(x, y)
-        assert output == pytest.approx(expected_mmd, abs=1e-6)
+            permuted_reference = Data(reference_data.data[::-1])
+            permuted_comparison = Data(comparison_data.data[::-1])
```
tm91236: I think you should be able to replace the if statement with 
```python
permuted_reference = reference_data[::-1]
permuted_comparison = comparison_data[::-1]
```
thanks to the `Data.__getitem__` method.

tests/unit/test_metrics.py
```diff
@@ -338,70 +325,55 @@ def _score_function(x: ArrayLike) -> Array:
...
-                Data(x.data), Data(y.data), laplace_correct=False, regularise=True
+            permuted_comparison = Data(
+                comparison_data.data[::-1], jnp.asarray(comparison_data.weights)[::-1]
             )
         else:
-            raise ValueError("Invalid mode parameterization")
-        # Compute the KSD using the metric object
-        assert output == pytest.approx(expected_ksd, abs=1e-6, rel=1e-3)
+            permuted_reference = Data(reference_data.data[::-1])
+            permuted_comparison = Data(comparison_data.data[::-1])
```
tm91236: Again, I think you can replace the if statement with
```python
permuted_reference = reference_data[::-1]
permuted_comparison = comparison_data[::-1]
```

tests/unit/test_metrics.py
tm91236, line 430: Would we ever expect the assertions on lines 375 and 376 to fail if the assertion on line 374 passes?

tm91236 (commented):
Thank you for the contribution! I've left a few minor comments, but seems to resolve #1140 as stated.

tests/unit/test_metrics.py
sylvesterkaczmarek: Updated, thanks. `Data.__getitem__` makes this much cleaner.

tests/unit/test_metrics.py
sylvesterkaczmarek: Updated here as well. Thanks for the suggestion.

tests/unit/test_metrics.py
sylvesterkaczmarek: Yes. The first assertion only checks permutation invariance against `expected`, which comes from the same KSD implementation on the unpermuted inputs. If both paths return the same invalid value, that comparison can still pass, so I would keep the finite and non-negative checks as independent KSD validity properties.

jd61349:
Thank you for this contribution! 

On reflection, there are a few other invariances that would be worth testing for. If you would like to add the following to this PR that would be great, if not, let us know and we can open a separate issue!

Invariances:
1. For translation invariant kernels the MMD is translation invariant, so if you add some constant c to all the data the result shouldn't change. 
2. If the kernel relies on the 2-norm then it is also invariant under orthogonal rotation. 
3. You should also be able to scale the input data, and the kernel lengthscale such that the result doesn't change. 

These invariances are really just properties of the kernel but it is good to test that the implementation of the MMD is sound.

sylvesterkaczmarek:
Added the three requested invariances in f8ece56: common translation, orthogonal rotation for the 2-norm kernel, and joint input/length-scale scaling. Each is covered for weighted and unweighted MMD. All 44 tests in test_metrics.py pass, and changed-file pre-commit passes.

jd61349 (approved):
Looks good, will merge when the commits are all verified @sylvesterkaczmarek. Thank you for the contribution!

sylvesterkaczmarek:
Done. I re-signed the three-commit series without changing the final tree and force-pushed it safely. GitHub now reports all three commits as Verified; the new head is `eee3afeeed`.

sylvesterkaczmarek:
@jd61349 All three commits are now GitHub Verified, as requested, and the full GitHub Actions matrix remains green. GitHub marks this PR APPROVED, CLEAN and mergeable. I have deliberately not rebased the 42-commit base drift because doing so would risk invalidating the existing approval/checks while GitHub does not currently require it for merge. This should now be ready for the merge you mentioned; if you specifically require a current-main rebase, I can do that instead.

PR #1237, opened 2026-09-10
fix: preserve Hermitian eigendecomposition semantics

### LINKED ISSUE #1236: Randomised eigensolver loses negative eigenvalues and mishandles zero and complex systems
Reproduced on `main` at `f954211` using the locked Python 3.12 dependencies.

```python
import jax.numpy as jnp
import jax.random as jr
from coreax.least_squares import RandomisedEigendecompositionSolver

solver = RandomisedEigendecompositionSolver(jr.key(1), oversampling_parameter=2)
identity = jnp.eye(2)
for matrix in [jnp.zeros((2, 2)), jnp.diag(jnp.array([-4., 2.])),
               jnp.array([[2., 1j], [-1j, 3.]])]:
    print(solver.solve(matrix, 0., identity, identity))
    print(jnp.linalg.pinv(matrix, hermitian=True))
```

The zero system returns NaNs instead of zeros. The indefinite diagonal system loses the `-0.25` inverse entry. The complex Hermitian system gives a non-Hermitian, incorrect inverse even though the sketch spans the entire matrix.

The inversion mask compares signed eigenvalues with the largest signed eigenvalue and includes exact zeros when the cutoff is zero. Range iteration, projection and reconstruction also use ordinary transposes where the documented Hermitian input requires conjugate transposes.

Expected: truncate by eigenvalue magnitude, preserve retained signs, exclude exact zero eigenvalues, and retain complex Hermitian structure. Full-rank sketches should agree with the Moore-Penrose inverse, including under JIT and batched solves.

I have deterministic reproductions and failing regression tests, and am preparing a focused fix.

sylvesterkaczmarek:
### PR Type

- Bugfix
- Tests

### Description

Fixes #1236.

Use eigenvalue magnitude for rank truncation while preserving eigenvalue signs, exclude exact zero eigenvalues, and use conjugate transposes throughout the Hermitian range finder, projection and reconstruction paths.

### How Has This Been Tested?

- Full unit suite: 1,780 passed, 19 skipped.
- Pre-commit checks passed, including Ruff, Pyright and Pylint.
- New tests cover negative-definite, indefinite, zero and complex Hermitian systems, JIT/non-JIT variants, low-rank reconstruction and batched solves.

No.

- [x] I have made corresponding documentation changes where appropriate.
- [x] There are no dependent downstream changes required.
- [x] I have updated CHANGELOG.md.


tests/unit/test_least_squares_spectral.py
jd61349, line 1: These tests should be incorporated into the existing `test_least_squares.py` file

jd61349 (changes requested):
Thank you for the contribution!

Looks good to me but the added tests should be incorporated into the existing test file.

sylvesterkaczmarek:
Moved the Hermitian regression cases into the existing `tests/unit/test_least_squares.py` in `04fbfe91` and removed the separate test module. The consolidated least-squares suite passes 52/52, and changed-file pre-commit passes.

jd61349:
I approved the workflows and it appears some tests are failing due to a tolerance issue.

sylvesterkaczmarek:
Fixed in 36fcb3f9. The failing case had ±0.5 exactly on the rcond cutoff, so backend-level eigendecomposition noise could move a boundary eigenvalue just above the threshold and produce a ±2 inverse entry. I moved the cutoff away from that boundary rather than relaxing numerical equality. The full least-squares file now passes 52/52 locally.

jd61349 (approved):
Looks great, thank you for the contribution!

sylvesterkaczmarek:
Thanks for the review and merge.

PR #1240, opened 2026-09-10
Add MMD particle gradient flow

### LINKED ISSUE #1105: Add gradient flow
### What's the new feature?

Subticket of #1104, add the gradient flow algorithm from https://arxiv.org/abs/1906.04370

### What value does this add?

Has been shown to outperform coresubset algorithms.

_No response_

### Additional context

_No response_

sylvesterkaczmarek:
### PR Type
Feature and tests.

### Description
Fixes #1105.

Add `GradientFlow` for pseudo-coresets using the MMD particle update from Arbel et al. (2019). Support weighted targets, deterministic flow, optional constant or scheduled gradient noise, and continued refinement with explicit solver state.

Particle updates now use a configurable `optax.GradientTransformation`. The default `optax.sgd(0.1)` preserves the original fixed-step behaviour, while adaptive optimisers such as Adam can be supplied. `GradientFlowState` preserves both Optax state and cumulative iteration so adaptive optimisers and noise schedules continue correctly across repeated `refine` calls.

Document the method, usage and limitations. Include analytic linear/Gaussian updates, independent noise checks, scheduled-noise continuation, adaptive-optimiser continuation, JIT/non-JIT execution, weight invariance, reproducibility, mixed precision and invalid-input tests.

### Validation
- 49 focused gradient-flow tests passed.
- Full unit suite: 1,789 passed, 19 skipped.
- Ruff, Pyright, cspell, Pyroma, formatting and changed-file checks pass.
- Pylint passes for the changed Python files when the repository's existing `suggestion-mode` option is omitted; local Pylint 4.0.7 otherwise rejects that upstream config option before linting.
- Sphinx HTML build passed with `-n -W`.
- GPU and full integration suites were not run.

### Compatibility
No new dependencies. Optax is already a Coreax dependency. Changelog and API documentation included.


coreax/solvers/gradient_flow.py
```diff
@@ -0,0 +1,204 @@
...
+    r"""
+    Move coreset points with maximum mean discrepancy gradient flow.
+
+    Implements the particle update in equation (21) of :cite:`arbel2019maximum`.
+    For current particles :math:`x_i`, normalised particle weights :math:`a_j`,
+    target points :math:`y_j` and normalised target weights :math:`b_j`, a step is
+
+    .. math::
+
+        x_i' = x_i - \gamma \left[
```
jd61349: Nitpick, use /prime here

coreax/solvers/gradient_flow.py
```diff
@@ -0,0 +1,204 @@
...
+    The kernel must be symmetric and differentiable. A suitable step size depends
+    on the kernel and data; arbitrary step sizes need not decrease the MMD.
+    Each iteration computes all particle-particle and particle-target gradients.
+
+    :param coreset_size: Number of particles, no larger than the target dataset
+    :param random_key: Key for initial sampling and optional gradient noise
+    :param kernel: Differentiable scalar-valued kernel
+    :param step_size: Non-negative finite Euler step size
+    :param num_iterations: Non-negative number of Euler steps per call
+    :param noise_scale: Non-negative finite gradient noise scale
```
jd61349: In the paper the noise scale can vary over time. Perhaps this could be implemented using https://optax.readthedocs.io/en/latest/api/optimizer_schedules.html

coreax/solvers/gradient_flow.py
```diff
@@ -0,0 +1,204 @@
...
+from typing_extensions import override
+
+from coreax.coreset import AbstractCoreset, PseudoCoreset
+from coreax.data import Data, SupervisedData
+from coreax.kernels import ScalarValuedKernel
+from coreax.solvers.base import ExplicitSizeSolver
+from coreax.util import KeyArrayLike
+
+
+class GradientFlowState(eqx.Module):
```
jd61349, line 35: This state could hold optimisation information such as the value of the loss, or the norm of the gradient etc.

coreax/solvers/gradient_flow.py
```diff
@@ -0,0 +1,204 @@
...
+                self.kernel.grad_x(evaluation_points, current_points),
+                initial.weights,
+            )
+            attraction = jnp.einsum(
+                "ijd,j->id",
+                self.kernel.grad_x(evaluation_points, target_points),
+                target.weights,
+            )
+            return (
+                current_points - self.step_size * (repulsion - attraction),
```
jd61349: While the paper does used a fixed step size. It is likely that performance would be improved using adaptive methods such as Adam.  Perhaps this should be reimplemented using Optax https://optax.readthedocs.io/en/latest/index.html, where the chosen optimiser is an input to the class.

jd61349 (changes requested):
Thank you for the contribution!

Before continuing the review, I think it would be a good idea to implement the optimisation loop using Optax to allow users to choose their optimiser, rather than relying on a fixed step size. 

The current implementation would then be subsumed by the use of optax.sgd(self.step_size)

coreax/solvers/gradient_flow.py
sylvesterkaczmarek: Updated in `c381b2e`: the equation now uses `x_i^\\prime`.

coreax/solvers/gradient_flow.py
sylvesterkaczmarek: Implemented in `c381b2e`: `noise_scale` can now be an Optax schedule evaluated at the cumulative optimisation step, and that iteration is preserved across `refine` calls.

coreax/solvers/gradient_flow.py
sylvesterkaczmarek: Updated in `c381b2e`: `GradientFlowState` now carries the Optax optimiser state and cumulative iteration, so adaptive optimisation and schedules continue rather than restarting.

coreax/solvers/gradient_flow.py
sylvesterkaczmarek: Reworked in `c381b2e`: `GradientFlow` now accepts an `optax.GradientTransformation`; the default `optax.sgd(0.1)` preserves the fixed-step behaviour, while adaptive optimisers such as Adam are supported and their state persists across refinements. Added a continuation regression for Adam.

sylvesterkaczmarek:
@jd61349 Addressed the review in `c381b2e`: configurable Optax optimiser, persisted optimiser state, scheduled noise with cumulative iteration, and the notation fix. 49 focused tests and the full 1,789-test unit suite pass; docs build is clean. Please take another look when convenient.

documentation/source/coreax/solvers.rst
```diff
@@ -4,3 +4,48 @@ Solvers
...
+from Arbel et al. The optimiser state and cumulative iteration are stored in
+``GradientFlowState``, so adaptive optimisers and scheduled noise continue correctly
+across repeated calls to ``refine``.
+
+``refine`` also accepts an existing coresubset or pseudo-coreset with exactly the
+configured number of points. Their weights and original data are preserved.
+Passing the returned state continues the optimiser, noise schedule and random sequence across calls.
+
+The implementation uses Euler steps for the MMD flow of :cite:`arbel2019maximum`,
+with a constant noise scale. Noise is applied to the locations at which the witness
```
jd61349: Noise scale is no longer constant.

tests/unit/test_gradient_flow.py
jd61349, line 1: I have no issues with these tests, they look good, though they should be incorporated into the tests of `test_solver.py`. We will likely want to split that file at some point as it is large, but for now these tests should live under `TestGradientFlow`, and inherit from the structure in there as much as possible.

jd61349 (commented):
One small doc issue, and the tests should be refactored under `test_solvers.py`!

documentation/source/coreax/solvers.rst
sylvesterkaczmarek: Updated in bdb4097f: the docs now describe the noise scale as configurable, either constant or scheduled.

tests/unit/test_gradient_flow.py
sylvesterkaczmarek: Moved the gradient-flow coverage into tests/unit/test_solvers.py under TestGradientFlow in bdb4097f, reusing ExplicitSizeSolverTest where its contract applies. The focused class now passes 57/57 tests.

jd61349:
It appears there is a small precommit issue @sylvesterkaczmarek

sylvesterkaczmarek:
Thanks for flagging that. Fixed in the latest commit: `TestGradientFlow.test_reduce` now keeps the base `SolverTest` signature/parametrization while retaining the gradient-flow-specific continuation semantics. The full pre-commit set for `tests/unit/test_solvers.py` is clean, including Pyright and Pylint, and `TestGradientFlow` passes 59/59.

jd61349 (approved):
Looks good!

jd61349:
Ah looks like one of your commits doesn't have the correct signature @sylvesterkaczmarek

sylvesterkaczmarek:
Done. I re-signed the four-commit series and force-pushed it without changing the final tree. GitHub now reports all four commits as Verified; the new head is e06f6224f5.

sylvesterkaczmarek:
Thanks for the review and merge, and for the useful Optax and test-structure feedback along the way.

PR #1241, opened 2026-09-10
fix(docs): preserve inherited method types and overloads

### LINKED ISSUE #795: Overloaded type annotations not compiled by Sphinx
### What's the problem?

We have used the `overload` decorator to provide more specific type hints in several places, e.g. in `kernels.ScalarValuedKernel`, but these are not rendered by Sphinx. Instead, Sphinx just shows the generic `Union` of several possible types. Moreover, if we rely on `overload` with class inheritance, no annotations are compiled at all, e.g. `kernels.scalar_valued.SquaredExponentialKernel.compute_elementwise`.

`overload` with class inheritance should work as of Python 3.12 - I'm unsure about earlier Python versions. Sphinx issues from a few years ago (pre-Python 3.12) also suggest this should work.

Compile documentation and look at classes mentioned above.

### Python version

3.12

### Package version

0.2.1

### Operating system

Windows

### Other packages

_No response_

### Relevant log output

_No response_

sylvesterkaczmarek:
Fixes #795.

Complete the type documentation for unannotated method overrides from compatible base methods. Preserve explicit subclass annotations and overloads, follow the method resolution order, and leave changed parameter lists alone. Runtime methods and numerical code are unchanged.

The current type-hint extension already renders directly declared overloads. This adds the missing inherited annotations and overloads for concrete kernels through a documentation-only hook.

Validation: 10 Sphinx regression tests passed, including real kernel methods and inheritance edge cases. Five fail without the new hook. The full HTML documentation build passed with `-E -n -W`; applicable pre-commit checks passed. Tested in the locked Python 3.12 environment.

The numerical unit suite was not rerun because no numerical code changed.


sylvesterkaczmarek:
Rebased onto current main and resolved the changelog conflict while preserving both upstream and #795 entries. The 10 documentation regressions pass, the full warning-clean Sphinx build passes, and the complete current pre-commit stack passes. I added only the narrowly scoped Pylint suppression required by Sphinx's six-argument callback signature, matching the existing pattern in `conf.py`. The branch is now one signed commit, 0 behind and mergeable. Fresh workflows are awaiting maintainer approval.

sylvesterkaczmarek:
@jd61349 This remains 0 behind and mergeable on the exact validated head. The 10 documentation regressions pass, the full warning-clean Sphinx build passes, and the complete current pre-commit stack passes. The only current GitHub-side blocker is maintainer approval of the fork-triggered Actions (Tests, Pre-commit, Documentation and Coverage), which are all sitting at action_required. Ready for review once those checks are allowed to run.

PR #1245, opened 2026-09-10
Restore the five-positional-argument lint limit

### LINKED ISSUE #782: Refactor methods with large numbers of positional arguments
### What's the issue?

As suggested by pylint's [R0917: too-many-positional-arguments](https://pylint.readthedocs.io/en/v3.3.0/user_guide/messages/refactor/too-many-positional-arguments.html), we should avoid having functions with a large number of positional arguments; functions which do should have a `*` inserted into their arguments list in order to mark later arguments as keyword-only.

This _is_ a breaking change, however, so we will have to be careful when doing this.

When doing this, reduce the allowed number of positional arguments in `.pylintrc` to the default of 5.

sylvesterkaczmarek:
Fixes #782.

Restore the global `max-positional-arguments` limit to 5 while keeping the existing function signatures unchanged, as requested in review. Existing signatures that legitimately exceed the limit now have narrow local Pylint exceptions, including the private solver helpers, established public callables, pytest-injected tests and the Sphinx event callback.

There are no runtime behaviour changes from `main`.

Validation:
- repository-wide `too-many-positional-arguments` scan: clean
- Ruff on all changed Python files: clean
- `git diff --check`: clean


tm91236 (changes requested):
Thank you for the contribution! Because these functions are all internally facing and are not part of the public API, we are happy to leave them as is. However, it would be sensible to drop the pylint arg limit back to the default of 5, and then add exceptions for this part of the internal api.

sylvesterkaczmarek:
@tm91236 Addressed in 04a79a3f: restored the existing helper signatures, removed the keyword-only regression tests and changelog entry, set the global positional-argument limit to 5, and added narrow local exceptions for the existing long signatures. Repository-wide R0917 scan, Ruff on changed Python files, and git diff --check are clean. I also updated the PR title and body to match the narrower scope.

tm91236 (approved):
Thank you for addressing the comments and your contribution. This looks good to merge.

tm91236:
Please can you make sure that all commits have verified signatures, 5b2ef08f5f7f5dea540ede4573b657cc4bfe2af7 appears to be unsigned.

sylvesterkaczmarek:
Thanks for spotting that. I rewrote the two-commit branch so both commits are SSH-signed and GitHub Verified. The final tree is unchanged; the new head is `4c66d31c`.

sylvesterkaczmarek:
Thanks for the review and merge. I appreciated the guidance to keep the internal helper signatures unchanged and narrow the lint change instead.

PR #1246, opened 2026-09-11
Replace flax usage with equinox

### LINKED ISSUE #582: Replace flax usage with equinox
### What's the new feature?

In `coreax.networks.py` we have the class `ScoreNetwork` which uses `flax.linen.Module`, as-well as the use of `flax.training.train_state.TrainState` in the same file. We could instead use Equinox functionality for the same purpose, and remove the need to use flax at all. Note that Equinox is used elsewhere in the codebase anyway. 

### What value does this add?

Reliance on fewer third party packages.

_No response_

### Additional context

_No response_

jd61349:
### PR Type

### Description

Replace all usage of Flax with Equinox. Fixes #582.

### How Has This Been Tested?

Related tests have been changed to reflect updated code, however what they test has remained the same. All tests pass.

No.


sylvesterkaczmarek (commented):
`MLP.__init__` now passes `mode="batch"` to `eqx.nn.BatchNorm`, but `pyproject.toml` still supports `equinox>=0.11.11`. In Equinox 0.11.11 `BatchNorm.__init__` has no `mode` argument, so a lowest-resolution install fails as soon as this benchmark constructs the model. Could we either raise the lower bound to the version that introduced this API or avoid the newer argument, and add a lowest-dependency benchmark smoke test?

jd61349:
> `MLP.__init__` now passes `mode="batch"` to `eqx.nn.BatchNorm`, but `pyproject.toml` still supports `equinox>=0.11.11`. In Equinox 0.11.11 `BatchNorm.__init__` has no `mode` argument, so a lowest-resolution install fails as soon as this benchmark constructs the model. Could we either raise the lower bound to the version that introduced this API or avoid the newer argument, and add a lowest-dependency benchmark smoke test?

Good spot! Thanks

sylvesterkaczmarek (approved):
Re-reviewed current `15a48daf`. The compatibility issue I raised is resolved: the Equinox lower bound is now `>=0.13.0`, matching the `BatchNorm(..., mode="batch")` API used by the benchmark, and the lowest/highest dependency test matrix is green across supported Python versions. No remaining blocker from me.

pyproject.toml
```diff
@@ -64,7 +63,7 @@ doc = [
 test = [
     {include-group = "benchmark"},
     "beartype>=0.19",
-    "pytest>=9",
+    "pytest >= 9, != 9.1, != 9.1.1",
```
tm91236: Please fix spacing and add a comment explaining why we need to exclude 9.1.0 and 9.1.1

pyproject.toml
```diff
@@ -30,8 +30,7 @@ dependencies = [
     # NOTE: lower bounds are the lowest supported version when using the 'lowest'
     # resolution mode of uv on the latest supported version of Python - the bounds
     # are only sharp for the latest supported Python version.
-    "equinox>=0.11.11",
-    "flax>=0.10.0",
+    "equinox>=0.13.0",
```
tm91236, line 33: This is a breaking change so should be added to the change log.

.pylintrc
```diff
@@ -107,18 +107,13 @@ recursive=no
 # source root.
 source-roots=
 
-# When enabled, pylint would attempt to guess common misconfiguration and emit
-# user-friendly hints instead of false-positive error messages.
-suggestion-mode=yes
-
```
tm91236, line 113: Why was this changed?

tm91236 (changes requested):
This looks good, but it does introduce some breaking changes without any deprecation warnings. I think we should keep the old network.py module, but mark the ScoreNetwork as deprecated for removal in v1.2.0.

.pylintrc
jd61349: Version 4.0 of Pylint removed it (https://pylint.readthedocs.io/en/latest/whatsnew/4/4.0/index.html):

"""

The suggestion-mode option was removed, as pylint now always emits user-friendly hints instead of false-positive error messages. You should remove it from your conf if it's defined.

Refs #9962 ([#9962](https://github.com/pylint-dev/pylint/issues/9962))

"""

We currently scope as `"pylint>=4.0.4",`.

pyproject.toml
jd61349: Added

pyproject.toml
jd61349: Fixed

sylvesterkaczmarek (changes requested):
Re-reviewed current `45e5a4b5`. Restoring `coreax/networks.py` for the requested deprecation path introduces a new compatibility problem: that module imports `flax`, but this PR still removes `flax` from the package dependencies. A normal installation can therefore no longer import the deprecated module that is being kept specifically to preserve compatibility. The current pre-commit run confirms this with unresolved imports for `flax`, `flax.linen`, and `flax.training` in `coreax/networks.py`. Please retain the `flax` dependency for the deprecation window (or otherwise make the compatibility module self-contained) and add a simple normal-install import regression for `coreax.networks`/`ScoreNetwork`.

documentation/source/conf.py
```diff
@@ -162,6 +166,8 @@
 nitpick_ignore = [
     ("py:class", "flax.core.scope.Scope"),
     ("py:class", "flax.linen.module._Sentinel"),
+    ("py:class", "flax.linen.module.Module"),
```
jd61349, line 169: docs were failing in workflow without these

jd61349:
> Re-reviewed current `45e5a4b5`. Restoring `coreax/networks.py` for the requested deprecation path introduces a new compatibility problem: that module imports `flax`, but this PR still removes `flax` from the package dependencies. A normal installation can therefore no longer import the deprecated module that is being kept specifically to preserve compatibility. The current pre-commit run confirms this with unresolved imports for `flax`, `flax.linen`, and `flax.training` in `coreax/networks.py`. Please retain the `flax` dependency for the deprecation window (or otherwise make the compatibility module self-contained) and add a simple normal-install import regression for `coreax.networks`/`ScoreNetwork`.

Agreed, thanks! Re-added the removed dependencies, and missing test file.

jd61349:
A note to whoever bumps to the next version, the following needs to be done wrt this PR:

1. Remove `flax`, `orbax-checkpoint`, and `tensorstore` dependencies.
2. Remove `networks.py` and `test_networks.py`
3. Ensure documentation files and nitpicks related to `flax` and `networks.py` are removed.

tests/unit/test_solvers.py
jd61349, line 1: Pre-commit was failing due to `pylint: disable-next=too-many-positional-arguments`

tests/unit/test_kernels.py
jd61349, line 1: Pre-commit was failing due to `pylint: disable-next=too-many-positional-arguments`

coreax/kernels/scalar_valued.py
jd61349, line 1: Pre-commit was failing due to `pylint: disable-next=too-many-positional-arguments`

sylvesterkaczmarek (approved):
Rechecked current 9b859ebd. The deprecated coreax.networks compatibility path is importable again in a normal install because flax is restored as a package dependency, and test_networks.py directly imports coreax.networks and exercises ScoreNetwork. The full lowest/highest dependency matrix is green. My compatibility blocker is resolved.

pyproject.toml
```diff
@@ -176,6 +176,8 @@ filterwarnings = [
     "error",
     "ignore::ImportWarning:umap",
     "ignore::DeprecationWarning:jaxopt",
+    "ignore:The 'coreax\\.networks' module is deprecated:DeprecationWarning",
+    "ignore:ScoreNetwork will be removed in version 1\\.2\\.0\\.:DeprecationWarning",
```
tm91236: Can this be `ignore:DeprecationWarning:coreax`?

tm91236 (changes requested):
Please can we remove the tests for the now deprecated ScoreNetwork, and add a new tests to check the deprecation warning is raised when attempting to instantiate the ScoreNetwork.

sylvesterkaczmarek:
One concrete follow-up on current 7eb669b7: the requested removal of the deprecated ScoreNetwork behavior tests has made the coverage gate fail. CI now reports coreax/networks.py at 54% and overall coverage at 97.27148704%, down from the 97.94% baseline by 0.666%, exceeding the repository tolerance. The functional/dependency matrix is green; this is specifically the coverage policy reacting to the intentionally untested deprecated module. That will need reconciling before merge, for example by whatever coverage treatment the maintainers prefer for code retained only for the deprecation window.

tm91236 (approved):
Thank you for making these changes. Reduction in code coverage is acceptable given the uncovered code is now deprecated.

PR #1252, opened 2026-09-12
fix: report accurate UTF-8 hook line numbers

### LINKED ISSUE #1251: coreax-metadata UTF-8 hook reports the wrong line number
The `coreax-metadata` repository's `require_utf8.py` pre-commit hook can report the wrong line number for invalid UTF-8 that occurs after the first line.

For example, a file with valid lines 1-2 and an invalid byte on line 3 currently reports the decode failure as line 1. `TextIOWrapper.readline()` may decode a larger internal buffer before returning the first line, so the exception location is not tied to the loop's current line counter.

Reading the file in binary mode and decoding each byte line independently preserves the actual line number and can report more than one invalid line in a single run.

sylvesterkaczmarek:
Fixes #1251.

Read files as binary lines and decode each line independently so invalid UTF-8 is reported against the correct physical line. The hook now also continues after a decode failure and reports multiple invalid lines in one run.

Adds regression coverage for invalid bytes on multiple lines and for valid UTF-8 input, and records the fix in the changelog.

Validation:
- `pytest tests/unit/test_require_utf8.py -q`: 2 passed
- pre-commit checks on all changed files: passed
- `git diff --check`: passed


sylvesterkaczmarek:
Rebased cleanly onto current main as one SSH-verified commit. The UTF-8 regression tests pass 2/2, the complete pre-commit suite passes, and the full current unit suite passes 1,835 tests with 19 skipped. The branch is now 0 behind and mergeable. Fresh workflows are awaiting maintainer approval.

sylvesterkaczmarek:
@jd61349 This remains 0 behind and mergeable on the exact validated head. The UTF-8 regression tests pass 2/2, the full unit suite passes 1,835 tests with 19 skipped, and the complete pre-commit suite passed. The remaining GitHub-side blocker is maintainer approval of the fork-triggered workflows. Ready for review once those checks are allowed to run.

